PR #14, opened 2020-09-23
Add reported but unmodelled benefits

nikhilwoodruff:
This adds input variables for all other weekly benefits in the FRS, such as Severe Disability Allowance, Maternity Allowance and more bespoke benefits. Also, the linting has been set at 79 characters per line.

### DIFF
--- a/openfisca_uk/reforms/basic_income/simulation_4.py
+++ b/openfisca_uk/reforms/basic_income/simulation_4.py
@@ -50,16 +50,33 @@ class family_net_income(Variable):
     definition_period = ETERNITY
 
     def formula(family, period, parameters):
+        benefits = [
+            "family_basic_income",
+            "child_tax_credit",
+            "working_tax_credit",
+            "child_benefit",
+            "income_support",
+            "housing_benefit_actual",
+            "contributory_JSA",
+            "income_JSA",
+            "DLA_SC_actual",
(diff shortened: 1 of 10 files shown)


nikhilwoodruff:
Also added poverty measures (absolute and relative) with references.

openfisca_uk/reforms/basic_income/simulation_4.py
```diff
@@ -50,16 +50,33 @@ class family_net_income(Variable):
     definition_period = ETERNITY
 
     def formula(family, period, parameters):
+        benefits = [
```
MaxGhenis, line 53:
```suggestion
        BENEFITS = [
```

MaxGhenis: https://www.python.org/dev/peps/pep-0008/#constants

MaxGhenis, line 53: Could `family_net_income` and `total_benefit_value` live somewhere outside the reform directories to be more accessible?

nikhilwoodruff: Ah yeah, that's definitely a good solution - was thinking of something similar because with the current way is a bit inelegant to redefine in reforms.

PR #17, opened 2020-09-26
Benefits modelling overhaul

nikhilwoodruff:
# Summary

This is a pretty substantial change in the handling of benefits by OpenFisca-UK, and the details are hopefully clear below.

## Benefits modelling

 Previously, we've taken the earnings data and started to add individual benefits in isolation, starting from reported amounts and eventually using completely simulated amounts. But in order to correctly measure poverty within the system, we need to take into account _all_ benefits and income, including uncommon benefits and the value of in-kind benefits such as free school meals and odd jobs. The FRS has a few aggregate variables containing these, so these changes shift from trying to build up pieces of the tax-benefit system in isolation to taking the aggregate variables (e.g. total disability benefit income) and then swapping out reported benefits for simulated benefits by subtracting the reported amount for e.g. Income Support and adding the computed Income Support amount back in. Of the benefits we have modelled so far, I haven't substantially changed the implementations apart from to be compatible with other changes and they all still pass the tests.

### Family -> BenUnit

All mentions to the Family entity - this is now renamed to the Benefit Unit to remove any ambiguity.

### Household

We now have a functional household entity, with variables for equivalised household net income (useful for poverty calculation) and poverty variables with helper variables giving the number of some key person types (e.g. adults, children) in the household.

## Organisation

Previously there wasn't really much thought into organising the variables, and there was a lot of overlap between variables in separate files e.g. taxes and benefits with variables like net income, so I've organised all the variables old and new by the entity they apply to, and standardised within that, for example entity variables might describe attributes of the entity (e.g. Person > Attributes > age, is head of benunit) or the entity's income, or implementation of the entity's benefits.


.vscode/snippets/python.json
```diff
@@ -0,0 +1,36 @@
...
+	// "Print to console": {
+	// 	"prefix": "log",
+	// 	"body": [
+	// 		"console.log('$1');",
+	// 		"$2"
+	// 	],
+	// 	"description": "Log output to console"
+	// }
+	"OpenFisca Variable": {
+		 	"prefix": "var",
```
MaxGhenis, line 16: quite the tabbing here

openfisca_uk/reforms/basic_income/reform_1.py
```diff
@@ -59,63 +64,52 @@ def formula(person, period, parameters):
             person("age", period) < 65
         )
         return (
-            person("is_senior", period) * 50
-            + adult_young * 55
-            + adult_old * 65
-            + person("is_child", period) * 22
+            person("is_senior", period) * 40
```
MaxGhenis, line 67: At some point it'd be good to have these as parameters, in the meantime could consider making them their own constants for more clarity, e.g. 
```
UBI_SENIOR = 40
...
return (
    person("is_senior", period) * UBI_SENIOR
...
```

openfisca_uk/reforms/basic_income/reform_3.py
```diff
@@ -0,0 +1,72 @@
...
+        for added_var in [basic_income, benunit_basic_income]:
+            self.add_variable(added_var)
+        for removed_var in [
+            "child_benefit",
+            "income_support",
+            "child_tax_credit",
+            "working_tax_credit",
+            "state_pension",
+        ]:
+            self.neutralize_variable(removed_var)
```
MaxGhenis, line 72: taking a list of variables in `neutralize_variable()` could be a helpful openfisca feature

openfisca_uk/variables/benunit/benefits/computed.py
```diff
@@ -0,0 +1,348 @@
...
+class income_support(Variable):
+    value_type = float
+    entity = BenUnit
+    label = u"Income Support amount received per week"
+    definition_period = ETERNITY
+
+    def formula(benunit, period, parameters):
+        younger_age = benunit("younger_adult_age", period)
+        older_age = benunit("older_adult_age", period)
+        personal_allowance = (
```
MaxGhenis, line 15: should this be its own concept? and/or could you describe the logic in a comment, given it's complexity?

openfisca_uk/variables/benunit/benefits/computed.py
```diff
@@ -0,0 +1,348 @@
...
+            + benunit("is_couple", period)
+            * parameters(
+                period
+            ).benefits.income_support.income_disregard_couple
+            + benunit("is_lone_parent", period)
+            * parameters(period).benefits.income_support.income_disregard_lone,
+        )
+        has_children = benunit.nb_persons(BenUnit.CHILD) > 0
+        random = np.random.rand(*has_children.shape)
+        takeup = (random < 0.89) * has_children + (random < 0.82) * (
```
MaxGhenis, line 70: could these be parameters, or made their own constants with descriptions?

openfisca_uk/variables/benunit/benefits/computed.py
```diff
@@ -0,0 +1,348 @@
...
+    definition_period = ETERNITY
+
+    def formula(benunit, period, parameters):
+        num_exempt_children = benunit.sum(
+            benunit.members("is_CTC_child_limit_exempt", period)
+        )
+        non_exempt_children = (
+            benunit.nb_persons(BenUnit.CHILD) - num_exempt_children
+        )
+        spaces_left = max_(0, 2 - num_exempt_children)
```
MaxGhenis, line 113: make 2 a parameter or constant

openfisca_uk/variables/benunit/benefits/computed.py
```diff
@@ -0,0 +1,348 @@
...
+class JSA_contributory(Variable):
+    value_type = float
+    entity = BenUnit
+    label = u"JSA (contributory) amount received per week"
+    definition_period = ETERNITY
+
+    def formula(benunit, period, parameters):
+        age = benunit("younger_adult_age", period)
+        is_couple = benunit.nb_persons(BenUnit.ADULT) == 2
+        single_young = (age >= 18) * (age < 25) * (np.logical_not(is_couple))
```
MaxGhenis, line 267: could 18 and 25 be params or constants

openfisca_uk/variables/benunit/benefits/computed.py
```diff
@@ -0,0 +1,348 @@
...
+
+class JSA_contributory(Variable):
+    value_type = float
+    entity = BenUnit
+    label = u"JSA (contributory) amount received per week"
+    definition_period = ETERNITY
+
+    def formula(benunit, period, parameters):
+        age = benunit("younger_adult_age", period)
+        is_couple = benunit.nb_persons(BenUnit.ADULT) == 2
```
MaxGhenis, line 266: is this a sufficient condition for this? what if an elderly parent is in the benunit?

openfisca_uk/variables/benunit/benefits/computed.py
```diff
@@ -0,0 +1,348 @@
...
+        )
+        threshold = (
+            eligible_for_both
+            * parameters(period).benefits.working_tax_credit.income_threshold
+            + (child_tax_credit_amount > 0)
+            * (1 - eligible_for_both)
+            * parameters(period).benefits.child_tax_credit.income_threshold
+        )
+        reduction = (
+            max_(0, (benunit("benunit_income", period) * 52 - threshold))
```
MaxGhenis, line 149: is it possible to normalize all data initially such that 52 doesn't have to show up in any logic code?

MaxGhenis (commented):
some ideas for future PRs

openfisca_uk/variables/benunit/benefits/computed.py
nikhilwoodruff: Ok, so there's a mistake here in the contributory JSA in that it seems it actually doesn't have the different rates - will change this. But I think the is_couple condition (if I understand the question right) is correct - a benefit unit can only be a single person or a couple, plus any dependent children so the benefit unit is a couple if and only if the number of adults is exactly 2.

nikhilwoodruff: This kind of ambiguity between "family" and "benefit unit" (nuclear family) is why I changed the references to family to benunit - hopefully that clears it up a bit.

openfisca_uk/variables/benunit/benefits/computed.py
nikhilwoodruff: We could, but the vast majority of benefits and NI rate specification is in terms of weekly income, so we'd either need to divide by 52 somewhere else, or provide both weeklyised and yearlyised income amounts?

nikhilwoodruff:
Thanks for these - apart from a few specific points here (will have another look at contributory JSA and Income Support), it seems like we need a big parameterising PR that eliminates most of the constants in the calculation logic. I also like the idea of the OpenFisca neutralise_variables function taking a list, maybe that could extend to other reform methods like add_variable.

openfisca_uk/variables/benunit/benefits/computed.py
nikhilwoodruff: I think in that case yes, you'd have:

- Benefit unit: 2 parents and the 16-year old dependent child
- Benefit unit: 25-year old child
- Benefit unit: elderly person

Issue #18, opened 2020-09-29
Account for net income disparity


nikhilwoodruff:
Currently the FRS survey data has adults with net incomes unexplainable from the survey data and current tax system alone (very low tax burdens, unexplainably high tax burdens too). One way to deal with this might be to measure the initial error in the baseline simulation and use this as an adjustment factor.

Issue #15, opened 2020-09-23
Test poverty measures


nikhilwoodruff:
The model now has variables for absolute and relative poverty, with references to government reports in #14, but testing these is probably necessary, which can be done with poverty line calculators such as [this one](https://www.childrenssociety.org.uk/what-we-do/policy-and-lobbying/child-poverty/poverty-line-calculator) to see if they're following the same method/parameters.

PR #21, opened 2020-10-06
Validate net income

nikhilwoodruff:
# Re-implement net income to match official figures

Long post, so TL;DR: aiming to better match official figures, mostly successful.

## In search of HBAI figures

Referencing #18 , and kind of #15.

This follows the general aim of trying to ensure that the model poverty rates and other aggregate figures reasonably match other official figures, such as the HBAI poverty rates. It turns out that the HBAI seems to follow through their own specific process for computing net income with some arbitrary choices about what's included and what's not. So in this PR I've tried to follow how I think they got their figures, and that includes the following changes:

- My last PR switched from building up gross income component-by-component in favour of using the FRS aggregate variables. I've flip-flopped on this because I started looking closer at the steps HBAI uses to generate their figures. Only using FRS "net income" variables does not reproduce the HBAI household net income figures, they have a ground-up process that I've tried to re-implement by using the relational tables to build variables from the tables:
  - accounts
  - adult
  - assets
  - benefits
  - benunit
  - child
  - external child maintenance expenses
  - household
  - job
  - pension
- Instead of storing reported benefits at the benefit unit level, store them at the person level. This is to try and make the income tax calculations as accurate as possible, because some benefits (e.g. Child Benefit after an income threshold, contributory JSA, and others) are taxed and we should make sure we're taxing the right people in the benefit unit. Simulated benefits are still calculated per benefit unit as before, but before taxation the burden is passed onto the head of the benefit unit. A future PR could make sure that they're passed to the non-head where this is the case, but I don't think this happens very often.
- Add equivalisation for AHC households. In some fine print that I missed before, the HBAI actually uses a slightly different equivalisation scheme when comparing household net incomes after housing costs are deducted. This has been added and is used by the poverty variables.
- Include a constant adjustment for when our net income differs a lot from the FRS reported net income (note. the FRS "net income variable" IS NOT the one HBAI uses, they calculate their own based on their own list of included/excluded components (e.g. maintenance expenses included/families with absent spouses excluded).

The result of this is that we get the following results for absolute poverty:

- Adults BHC: 17% (HBAI 15%)
- Adults AHC: 20% (HBAI 20%)
- Children BHC: 18% (HBAI 17%)
- Children AHC: 25% (HBAI 26%)

Which is pretty close, given that we're using the tax rules of 2 years later and generally simulating income taxes rather than using the actual reported amounts.

As usual, I've adjusted the reforms slightly to keep them budget-neutral.  They all now show the medium reductions in adult poverty and stronger reductions in child poverty, which is more familiar. Of course we should be expecting what we aim for rather than aiming for what we expect, but it seems encouraging that the model isn't completely in a world of its own.

## General housekeeping

There's also some changes in the ongoing effort to clean up the division of tasks between OpenFisca-UK and the static analysis repo. At the moment we generate the input dataset using the analysis repo, and have some helper functions in there to handle OpenFisca-UK output. I've moved the dataset generation out of the analysis repo and into a new repo [frs-tools](https://github.com/nikhilwoodruff/frs-tools), and have started moving some of the OpenFisca-UK helper functions into OpenFisca-UK itself so they're easily importable. Not sure whether this breaks any OpenFisca ethos but seemed neater. The changes:
- Remove the FRS variable holders (OpenFisca-UK currently takes in a few internal FRS variables and calculates its own variables from them). This split a task between this and the dataset generation script in a way that wasn't very neat, so this has been moved onto the new repository [frs-tools](https://github.com/nikhilwoodruff/frs-tools) which generates the input dataset.
- Add the easy model and dataframe generator functions into OpenFisca-UK.


nikhilwoodruff:
One other thing I neglected to mention - I've quarantined Working and Child Tax Credit because they seemed to be one of the main reasons that the poverty figures were previously so out from the HBAI figures, but I can't seem to figure out why right now. In terms of UBI analysis this just means that we'll have to not include UBI in the means tests for them, leading to about £15bn (allegedly) lost revenue from marginally low-income families.

PR #24, opened 2020-10-17
Benefits revamp

nikhilwoodruff:
# Return of the Benefits

After finding issues with the Tax Credits amounts, I went back and followed the [UKMOD country report](https://www.iser.essex.ac.uk/files/projects/UKMOD/EUROMOD_country_report.pdf) implementations of the core benefits as closely as possible. We now have modelled:

- JSA (contribution-based) - re-implemented, correctly at the person-level this time
- JSA (income-based) - re-implemented
- Working Tax Credit - minor corrections (aggregate issues still persist)
- Child Tax Credit - minor corrections
- Pension Credit - new
- Income Support - re-implemented
- Housing Benefit - new
- Universal Credit - new
- Winter Fuel Allowance - new, very minor

Most of these benefits give pretty close total expenditure figures when masked with the reported amount being above zero, though there often seems to be some inaccuracy in the tens of pounds per week with a few benefits, although given that Tax Credits are based on the previous tax year's income, which we are estimating from one weeklyised payment this year, I'm not sure of the consequence of this.

UBI reforms show that these benefits seems to respond appropriately when we means-test them away, and give good poverty reductions, amid some issues with pensioner poverty - pensioners are treated pretty favourably by the UK tax system, so our flat tax models can be quite punitive on them.


nikhilwoodruff:
Just adding this for a bit more detail on the validity of the benefits - they are all restricted to prior claimants to try and keep comparability with official HBAI figures which as I understand take the reported amounts. We get the following total expenditure per year figures:
(the actual figures are taken from 2018/19 forecasts, and are always close to 2017/18 actual figures)
- WFA: reported: £2bn, simulated £3bn, actual £2bn (therefore not simulated in reforms at the moment, it's not means-tested)
- JSA (contribution-based): £120m, simulated £120m, actual £207m
- JSA (income-based): reported: £1.3bn, simulated £1.4bn, actual £1.4bn
- Income Support: reported: £2.1bn, simulated £1.9bn, actual £1.8bn
- Working Tax Credit + Child Tax Credit: reported £16.2bn, simulated £17bn, actual £22bn (forecast slightly different to other figures I previously looked at)
- Income Tax revenue: reported: unknown, difficulty consolidating, simulated: £180bn, actual £185bn
- NI (employee/Class 1) revenue: simulated £57bn unchecked so far, need to find figures which separate employee, employer 
 and other contribution classes
- Pension Credit: reported £3.1bn, simulated £3.6bn, actual £5.1bn
- Housing Benefit: reported £16bn, simulated £22bn, actual £21bn
- Universal Credit*: reported £5.4bn, simulated £6.3bn, actual £8.2bn

* Universal Credit is less tidy than the others with its long and complicated partial rollout: in 2017/18 the actual figure was £3.4bn, and the projected figure for 2018/19 was £8.2bn which is quite a jump.

Now there's an interesting phenomenon here which is that even with the eligibility restricted to current recipients, the simulated figures more closely match the official estimates, which suggests that often the reported amounts themselves are underestimates, as well as the issue of non-reporting. Or the model is overestimating benefit amounts and underestimating recipients in just the right amounts to get the official figures most of the time.

Issue #28, opened 2020-10-31
Add UBI classes for each demographic group


MaxGhenis:
e.g. `basic_income_u18`, `basic_income_pensioner`, etc.

This would make it so that these kinds of reforms wouldn't have to extract features about the person:

https://github.com/nikhilwoodruff/openfisca-uk/blob/e9139d401f03f8a48785049ce3c60fc5c2443746/openfisca_uk/reforms/basic_income/reform_1.py#L50-L54

MaxGhenis:
Fixed in #29

PR #29, opened 2020-11-08
Change handling of reforms

nikhilwoodruff:
# Reform handling

Up until now reforms have been manually programmed into OpenFisca-UK, and manually adjusted and reinstalled where needed. Here's what I propose specifically for the UBI reforms (as they are rapidly fine-tuning at this point):

- Take them out of the model (not 100% sure about this, we could probably leave them in after they're finalised)
- Include in OpenFisca-UK tools to **generate new reforms**. In this PR I've included a function which uses a simple binary search to solve for a budget-neutral UBI reform under given constraints and set ratios between child/adult/disabled child/etc.


nikhilwoodruff:
Also implements #28

PR #30, opened 2020-11-08
Update variables stats

nikhilwoodruff:
Hopefully the last PR before 0.1.0. This updates the statistical summaries of all variables in the simulation under 2018/19 FRS data.

### DIFF
--- a/docs/VARIABLE_STATS.md
+++ b/docs/VARIABLE_STATS.md
@@ -8,8 +8,8 @@ JSA_contrib_eligible
 
 adult_weight
   - description: FRS weighting of the person if they are an adult, 0 if a child (none provided by the FRS)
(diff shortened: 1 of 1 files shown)


MaxGhenis:
Is this generated from a script? Do UBI amounts belong here?

nikhilwoodruff:
Yeah so I took #28 to mean add the classes to the rest of the variables, rather than defining them in each reform, but not sure that's the way to go. Yep, the stats here are generated by a script, but the basic income variables I think you're talking about aren't the actual amounts, they're stats of the boolean variables ("(is a ) basic income pensioner"), e.g. basic_income_adult having a mean of 0.77 reflecting 77% of individuals in the FRS being over 18.

PR #41, opened 2020-11-26
Minor bug fixes

nikhilwoodruff:
Just some minor bug fixes that mean OpenFisca-UK now matches official poverty statistics.

- Fixed State Pension age calculation bug
- Restricted Pension Credit to current claimants as with other benefits
- Added maintenance payments and student loan repayment to expenses to match official poverty calculations (HBAI)

### DIFF
--- a/openfisca_uk/tools/optimisation.py
+++ b/openfisca_uk/tools/optimisation.py
@@ -78,7 +78,7 @@ def ft_funded_ubi_reform(
     class gross_income(Variable):
         value_type = float
         entity = Person
-        label = "Gross income"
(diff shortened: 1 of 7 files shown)


nikhilwoodruff:
I added the reporting requirement to the pension credit test, since we're requiring it as a condition of eligibility. Pension credit has one of the lowest take-up rates (~60%), so definitely needs the adjustment/restriction. One other issue this brings up is if we simulate a random take-up, we need to ensure the tests aren't flipping between pass/fail depending on the numpy random seed.

Issue #47, opened 2020-11-30
Create new disabled flag for UBI


MaxGhenis:
Should include current disabled flag, plus:
* income-based ESA
* registered disabled
* Equality Act flags
* PIP
* IIDB
https://github.com/PSLmodels/openfisca-uk/blob/86c76547670ba2d4b6c9b1b66f7139a44a098f02/openfisca_uk/variables/person/disability.py#L13-L20

NB: incapacity benefit and SDA have been rolled into ESA. Constant Attendance Allowance and New Style ESA are absent from FRS. Disability premiums are added into their respective benefits, so can't be disentangled to add to the logic.

nikhilwoodruff:
Also add PIP and ESA-income to disabled flag at the moment

PR #49, opened 2020-12-01
Disability

nikhilwoodruff:
Related to [the frs PR](https://github.com/nikhilwoodruff/frs/pull/7). This adds IIDB to the variables, and the Equality Act definition indicators, and a new ```is_standard_disabled``` variable. The ```is_standard_disabled``` variable is true for 18% of the population.


openfisca_uk/variables/person/disability.py
```diff
@@ -60,3 +60,38 @@ def formula(person, period, parameters):
...
+    def formula(person, period, parameters):
+        QUALIFYING = [
+            "is_disabled",
+            "is_severely_disabled",
+            "is_enhanced_disabled",
+            "IIDB",
+            "registered_disabled",
+            "dis_equality_act_core"
+        ]
+        return add(person, period, QUALIFYING, options=[MATCH]) + person.benunit("ESA_income", period, options=[MATCH]) > 0
```
MaxGhenis, line 97: Does this mean that all persons in a benefit unit with `ESA_income>0` is classified as disabled?

Also seems like this line might be too long, and lacks a newline.

MaxGhenis (commented):
Should this be linked to #47?

openfisca_uk/variables/person/disability.py
nikhilwoodruff: Good catch, I'll take it out

nikhilwoodruff: Good catch, I'll take it out

nikhilwoodruff:
Just added the black formatter check GitHub action again

nikhilwoodruff:
Also, removed the ```debug.log``` from the repo as raised in #50 , interestingly it didn't appear in my VSCode install, seems like it was generated and protected by a [bug](https://github.com/microsoft/vscode/issues/105191).

PR #69, opened 2020-12-23
Add ESA_income to list of qualifying benefits for base disability UBI

MaxGhenis:
Following @nikhilwoodruff's addition of `ESA_income` as a personal variable (given to an arbitrary member of the benefit unit) in #68.

### DIFF
--- a/openfisca_uk/variables/person/disability.py
+++ b/openfisca_uk/variables/person/disability.py
@@ -116,10 +116,10 @@ class is_disabled_for_ubi(Variable):
             "DLA_M",
             "DLA_SC",
             "IIDB",
-            # Given to a single person at benunit level
             "PIP_DL",
             "PIP_M",
-            "ESA_income",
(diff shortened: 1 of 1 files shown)


openfisca_uk/variables/person/disability.py
```diff
@@ -119,7 +119,7 @@ def formula(person, period, parameters):
             # Given to a single person at benunit level
             "PIP_DL",
             "PIP_M",
-            # ESA_income omitted since it is reported at benunit level
+            "ESA_income",
```
nikhilwoodruff, line 122:
```suggestion
            "ESA_income_reported_personal",
```

In the PR (#68)  adding the personal ESA income benefit, I kept the original ```ESA_income```, and just added the personal reported amount, ```ESA_income_reported_personal```, so we'd probably want to use this?

nikhilwoodruff (changes requested):

PR #70, opened 2020-12-31
Use reported benefits in disability imputation

nikhilwoodruff:
Make sure we're using reported benefits, which won't be neutralised.


openfisca_uk/variables/person/disability.py
```diff
@@ -109,17 +113,17 @@ class is_disabled_for_ubi(Variable):
...
+            "incapacity_benefit_reported",
+            "SDA_reported",
+            "AA_reported",
+            "DLA_M_reported",
+            "DLA_SC_reported",
+            "IIDB_reported",
+            # Given to a single person at benunit level
+            "PIP_DL_reported",
+            "PIP_M_reported",
+            # ESA_income omitted since it is reported at benunit level
```
MaxGhenis, line 126: We can add this now right?

MaxGhenis (commented):
.

PR #74, opened 2021-01-07
Remove is_{severely,enhanced}_disabled_for_ubi and add ESA_income as qualifying UBI disability

MaxGhenis:


MaxGhenis:
The `ESA_income_reported` didn't work, got this error:
```
ValueError: You tried to compute the variable 'ESA_income_reported' for the entity 'people';
however the variable 'ESA_income_reported' is defined for 'benunits'.
```

nikhilwoodruff (approved):

MaxGhenis:
Works now after changing to `ESA_income_reported_personal`

Would like to add a test before merging though

nikhilwoodruff:
Thanks @MaxGhenis - I've just added an example test for the disability flag, hopefully it's easy to follow. Right now the tests can be slightly difficult to specify when periods are different - for example, benefits defined on a per-week level are used to define yearly flags. It's not an issue for this, because the benefit receipt just needs to be positive, but if we were thresholding we'd need to specify the receipt in each week of the year. To solve this, I think we just need to use the ```set_input``` field that we haven't so far ([reference](https://openfisca.org/doc/coding-the-legislation/35_periods.html#automatically-process-variable-inputs-defined-for-periods-not-matching-the-definition-period)).

MaxGhenis:
Oh cool thanks @nikhilwoodruff! Merging.

PR #76, opened 2021-01-12
Add age_under_18, age_18_64, age_over_64

### LINKED ISSUE #75: Add age flags
0-17, 18-64, and 65+

So we can sum them up by household

MaxGhenis:
Fixes #75

I think it'd also be clearer to rename other purely-age-based flags like `is_young_child` with the `age_{lower}_{upper}` pattern (or remove them if they're not used). e.g. right now `age_0_17` is identical to `is_child` for backwards-compatibility, but especially given FRS has a separate child flag I think `is_child` would be worth removing.

### DIFF
--- a/openfisca_uk/variables/person/attributes.py
+++ b/openfisca_uk/variables/person/attributes.py
@@ -154,10 +154,10 @@ class is_young_child(Variable):
         return person("age", period.this_year) < 14
 
 
-class age_0_17(Variable):
+class age_under_18(Variable):
     value_type = bool
     entity = Person
-    label = u"Whether the person is age 0 to 17"
+    label = u"Whether the person is under age 18"
     definition_period = YEAR
 
     def formula(person, period, parameters):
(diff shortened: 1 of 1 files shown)


nikhilwoodruff (commented):
Looks good - though minor aesthetic suggestion is to have age_0_17 be age_under_18 if we're going with age_65_plus?

openfisca_uk/variables/person/attributes.py
```diff
@@ -154,6 +154,37 @@ def formula(person, period, parameters):
         return person("age", period.this_year) < 14
 
 
+class age_0_17(Variable):
```
nikhilwoodruff, line 157:
```suggestion
class age_under_18(Variable):
```

MaxGhenis:
How about `age_under_18` and `age_over_64` for more consistency? Now all intervalled ones are `<=`/`>=` and open-ended ones are `<`/`>`.

PR #83, opened 2021-01-27
Frs update

nikhilwoodruff:
# Handle changes in ```frs```

This adds a few changes, some reflecting those made in the frs package:

- Be more consistent with reporting variables
  - Previously, "_reported" variables would be at different entity levels, and may be weekly but with YEAR periods. This makes them all person-level and yearly values. For reporting variables "x" previously at benunit level, there is now a variable "benunit_x".
- Optimise the map_to function, and add more options for disambiguation
  - E.g. instead of just map_to with unexplained logic, we have copy_to_person, average_by, etc.
- Add code for SurveySimulation backend, not yet in use
- Better consistency in input variables: assume yearly, except when a "weekly" or other period is prefixed. E.g. "hours" is the yearly amount of hours, but "weekly_hours" is the average weekly hours.

Test have only been changed to be valid (variables set for correct entity levels, uprating from weekly to yearly amounts where needed).

@MaxGhenis this goes with the frs PR: I'll double check it doesn't break the UK UBI code before merging.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setup(
         "pytest==5.4.3",
         "OpenFisca-Core @ git+https://github.com/nikhilwoodruff/openfisca-core",
         "openfisca-survey-manager",
-        "frs==0.2.0",
+        "frs @ git+https://github.com/nikhilwoodruff/frs",
         "microdf @ git+https://github.com/pslmodels/microdf",
         "rdbl",
     ],


MaxGhenis:
Is `map_to` in `openfisca-core`? I'm curious how much of this could plausibly move upstream.

nikhilwoodruff:
Closing, to re-add in pieces.

PR #90, opened 2021-02-17
Remove ESA as a UBI-supplement-qualifying program

MaxGhenis:
Since it's heavily means-tested, and results are still more progressive without it.


nikhilwoodruff (commented):
Looks good, thanks @MaxGhenis

PR #110, opened 2021-04-25
Zero-out unimputable council tax records

nikhilwoodruff:
Very minor bug-fix (introduced by the data-handling PR): I noticed that where we impute missing council tax values (around 80% are missing) stochastically by region-council tax band intersection, around 3% are NaNs. Council tax informs household net income, so can mess up e.g. decile-level averages. I've zeroed out council tax for now for those 3% until a better solution exists.

### DIFF
--- a/openfisca_uk/microdata/frs/dataset.py
+++ b/openfisca_uk/microdata/frs/dataset.py
@@ -167,6 +167,7 @@ class FRSDataset:
         household.H_CTANNUAL = np.where(
             household.H_CTANNUAL == 0, ct, household.H_CTANNUAL
         )
+        household = household.fillna(0)
 
         # store dataset for future use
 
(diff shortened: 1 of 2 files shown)


openfisca_uk/microdata/frs/dataset.py
```diff
@@ -167,6 +167,7 @@ def generate_dataset(self, year: int) -> None:
         household.H_CTANNUAL = np.where(
             household.H_CTANNUAL == 0, ct, household.H_CTANNUAL
         )
+        household = household.fillna(0)
```
MaxGhenis, line 170:
```suggestion
        household.fillna(0, inplace=True)
```

MaxGhenis: (if it's a Series where this works)

MaxGhenis (commented):
Is this 3% NaN because there are no observations for their region-council tax band? How about using the overall average instead of zero?

PR #127, opened 2021-05-26
Docs

nikhilwoodruff:
This updates the main documentation to include up-to-date information on weighting, as well as more helpful examples for both parametric and structural reforms.


MaxGhenis:
You shouldn't needs `docs/book/.ipynb_checkpoints` and `docs/book/_build` (that'll get built automatically from jb). I'd suggest removing them and adding them to `.gitignore`.

nikhilwoodruff:
Until we can find a way to appropriately make the non-public datasets available to GitHub Actions, the notebooks are run locally, which is why the files are included

PR #131, opened 2021-06-04
Have parity between IndividualSim and Microsimulation reform handling

nikhilwoodruff:
This just adds the functionality for IndividualSim objects to handle reforms in the same way as Microsimulation objects (as in, tuples of reforms are reforms, applied in their order).


MaxGhenis:
(This PR also removes the built html jupyter-book files)

PR #134, opened 2021-06-21
Add LHA rates

nikhilwoodruff:
This PR adds the LHA rates for financial years 2018-2022 to the model. The default rates are unchanged: previously they were hard-ccoded as Maidstone, Kent rates, now there is a `BRMA` variable with a default value of `MAIDSTONE`. To fetch the LHA rate, the model looks three things in the parameter tree: the BRMA, the LHA category (shared, 1 bed, ..., 4+ beds) and the current financial year. The LHA category still takes into account LHA rules on acceptable number of bedrooms (e.g. room-sharing policies).

@MaxGhenis FYI, this was one of the last big features that benefits calculators all have that the model didn't. The next one is actual Council Tax rates by Local Authority.


MaxGhenis:
Thanks @nikhilwoodruff, and it's good to see here how openfisca yaml files accept inputs beyond year. This would provide a closer analog to taxcalc, which uses lists for many tax variables when they vary by marital status. When do you think elements of the tax+benefit system should use this, e.g. could/should UC allowances be made a single structured file by age and marital status rather than [the current four files](https://github.com/PSLmodels/openfisca-uk/tree/master/openfisca_uk/parameters/benefit/universal_credit/allowances)?

nikhilwoodruff:
Thanks, and yeah I think those UC allowances could be one file. Though since the folder > subfiles approach is equivalent to the yaml object: child syntax, I think a good rule of thumb is to separate files such that each file has one description value, and any child objects are clear in meaning.

openfisca_uk/variables/finance/benefit/family/LHA.py
```diff
@@ -76,12 +92,42 @@ def formula(benunit, period, parameters):
...
+    value_type = float
+    entity = BenUnit
+    label = u"LHA Rate"
+    definition_period = WEEK
+
+    def formula(benunit, period, parameters):
+        personal_BRMA = benunit.members.household(
+            "BRMA", period.this_year
+        ).decode_to_str()
+        BRMA = (
```
MaxGhenis, line 115: Am I interpreting this right as selecting the BRMA of the first member of the benefit unit? Could you add a comment on this?

MaxGhenis (approved):

openfisca_uk/variables/finance/benefit/family/LHA.py
nikhilwoodruff: Yeah, this was a bit complicated. Essentially, we need to transpose a household's BRMA onto each benefit unit because housing benefit is at the family level, rather than household level. Usually we can just project to person-level (person_BRMA = `benunit.members.household("BRMA")`) and then aggregate to benefit unit level (`family_BRMA = benunit.value_from_first_person(person_BRMA)`). But this doesn't work for the Enum type, which gets stuck as the number, and the decode function is lost somehow.

MaxGhenis: Gotcha, could you add a comment to the code? Should there be an issue in openfisca-uk or openfisca-core for the projecting and aggregating enums?

MaxGhenis: Gotcha, could you add a comment to the code? Should there be an issue in openfisca-uk or openfisca-core for the projecting and aggregating enums?

PR #135, opened 2021-06-25
Install family-resources-survey from github not pip

MaxGhenis:
Fixes an error


nikhilwoodruff:
Yep, the error being that some household net incomes were NaN in years 2019-onwards - caused by an old installation of family-resources-survey not handling earnings uprating, and therefore the model fell back silently to the toy dataset which didn't work for 2019.

PR #147, opened 2021-07-17
Microdata

nikhilwoodruff:
@MaxGhenis FYI

This PR updates the model to use the `openfisca-uk-data` package instead of separate `family-resources-survey`/`survey-personal-incomes` packages. This hopefully provides a more unified and easily extensible interface for using microdata. Currently in progress.

- [x] FRS input
- [x] SPI input
- [x] CLI
- [ ] WAS input (optional)
- [ ] FRS-SPI synthesis (optional)


nikhilwoodruff:
To note for future: there was a bug (fixed now) in which the poverty rate dropped from 16% to around 8% - this ended up being caused by the HD5 format saving string arrays as binary strings, which do not pass equality tests with the same text in Unicode (in the model, this caused everyone to default to being a renter, increasing Housing Benefit and Universal Credit artificially). Fixed now by adding a check in the data loading function.

MaxGhenis (approved):

PR #148, opened 2021-07-19
Add tests for microsimulation results

nikhilwoodruff:
This PR adds tests for the microsimulation model - on the poverty rate and the aggregate tax and benefit variables.

### DIFF
--- a/tests/with_synth/test_aggregates.py
+++ b/tests/with_synth/test_aggregates.py
@@ -1,13 +1,13 @@
-from openfisca_uk_data import SynthFRS
+from openfisca_data import SynthFRS
 from openfisca_uk import Microsimulation
 import pytest
 
 sim = Microsimulation(dataset=SynthFRS)
 
-average_tests = (("in_poverty_bhc", 0.15, 0.1),)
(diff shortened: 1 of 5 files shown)


tests/with_synth/test_aggregates.py
```diff
@@ -0,0 +1,33 @@
+from openfisca_uk_data import SynthFRS
+from openfisca_uk import Microsimulation
+import pytest
+
+sim = Microsimulation(dataset=SynthFRS)
+
+average_tests = (("in_poverty_bhc", 0.15, 0.1),)
```
MaxGhenis, line 7: capitalize as constant? also agg_tests

MaxGhenis, line 7: could you add a comment indicating that the 0.1 is the absolute window rather than relative? (also this seems high)

nikhilwoodruff: Yeah it is rather high unfortunately since the synthetic dataset isn't so close here. Shuffling fields doesn't affect individual variable aggregates like tax so much but does change household and family aggregates a fair amount.

Issue #171, opened 2021-08-28
Update policy parameters for 2021


MaxGhenis:
For example, OF sets Child Benefit at [21.05](https://github.com/PSLmodels/openfisca-uk/blob/master/openfisca_uk/parameters/benefit/child_benefit/amount/eldest.yaml) for eldest/only children, as of 2021-06-01, but it's currently [21.15](https://www.gov.uk/child-benefit/what-youll-get).

nikhilwoodruff:
Think we can close this.

PR #172, opened 2021-08-28
Update Child Benefit for 2021, historical data, and date of annual update

MaxGhenis:
* Minor change
* Impacted periods: all, but most importantly 2021+
* Impacted areas: `parameters/benefit/child_benefit/amount`
* Details:
  - Updates Child Benefit to its current value, which changed on 2021-04-06
  - Adds Child Benefit amounts for 2004-2009 per the [current source](https://revenuebenefits.org.uk/child-benefit/guidance/how-much-can-your-client-get/rates-and-tables/)
  - Edits the date of switch from YYYY-06-01 to YYYY-04-06 per [gov.uk](https://www.gov.uk/government/publications/rates-and-allowances-tax-credits-child-benefit-and-guardians-allowance/tax-credits-child-benefit-and-guardians-allowance)
  - Adds that gov.uk as a second reference separated by a space (PLMK if this is the correct way)

- Fix or improve an already existing calculation. See #171


nikhilwoodruff (approved):
Thanks @MaxGhenis for this. Though one thing to consider is, now that we're using yearly parameters, how best to handle mid-year parameter changes.

MaxGhenis:
Tests are breaking due to openfisca-data being split (#173), so merging

PR #184, opened 2021-08-30
Update NI and tax params for 2021

MaxGhenis:


openfisca_uk/parameters/tax/income_tax/allowances/personal_allowance/maximum_ANI.yaml
```diff
@@ -1,7 +1,7 @@
 description: Maximum adjusted net income before the PA is reduced.
 values:
-  2009-07-21:
-    value: 100000.00
+  2009-07-21:  # Date legislation passed. Is this when it was enacted?
```
nikhilwoodruff, line 3: Checked, this came into effect on 2016-04-06: from the notes, the amending Act (Finance Act 2015) specifies that the changes come into effect from the start of the 2016-17 tax year: https://www.legislation.gov.uk/ukpga/2015/11/section/5#section-5-4-a 
But this won't change any model outputs before the 2016-17 year.

nikhilwoodruff (approved):

PR #186, opened 2021-08-30
Add year information and update table.

### LINKED ISSUE #182: Describe time period in validation Jupyter Book page
https://pslmodels.github.io/openfisca-uk/validation looks great, but could you add what year OF, UKMOD, and the admin sources are from?

nikhilwoodruff:
Fixes #182


MaxGhenis (approved):
LGTM, could you remove the labour income breakdown? Seems like a useful set of totals to keep track of, but not here which focuses on validating against external sources.

nikhilwoodruff:
> LGTM, could you remove the labour income breakdown? Seems like a useful set of totals to keep track of, but not here which focuses on validating against external sources.

Done!

PR #191, opened 2021-09-01
Minor changes to be compatible with tax-benefit.org validators

eraviart:
* Technical improvement. | Minor change.
* Impacted periods: all.
* Impacted areas:
  * `openfisca_uk/__init__.py`
  * `openfisca_uk/parameters/`
* Details:
  - Add CURRENCY class attribute to class CountryTaxBenefitSystem.
  - Use `unit` instead of `threshold_unit` for parameters that are not of type Scale
  - Use 0001-01-01 instead of 0000-01-01

I suggest these changes, to be more compatible with other OpenFisca country packages (namely OpenFisca-France, OpenFisca-Senegal & OpenFisca-Tunisia).


nikhilwoodruff (approved):
Thanks for this @eraviart - looks good to me!

PR #233, opened 2021-09-27
Fix synthetic dataset loading

nikhilwoodruff:
This ensures that the default installation of openfisca-uk works with the synthetic dataset by default. Also removes the old synthetic.py file (this does not work any more, and predates the module in openfisca-uk-data).


nikhilwoodruff:
Closing to re-file from a branch.

PR #234, opened 2021-09-27
Ensure first installation without microdata runs

nikhilwoodruff:
This ensures that the default installation of openfisca-uk works with the synthetic dataset by default. Also removes the old synthetic.py file (this does not work any more, and predates the module in openfisca-uk-data).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ from setuptools import setup, find_packages
 
 setup(
     name="OpenFisca-UK",
-    version="0.4.3",
+    version="0.4.4",
     author="UBI Center",
     author_email="nikhil.woodruff@ubicenter.org",
     classifiers=[


openfisca_uk/initial_setup.py
```diff
@@ -48,6 +49,9 @@ def main():
                 print(
                     f"You currently have the synthetic FRS dataset for: {', '.join(map(str, SynthFRS.years))}."
                 )
+            with open(REPO / "tools" / "datasets.yml") as f:
```
MaxGhenis, line 52: Is this so that it updates previous installations?

MaxGhenis (approved):

openfisca_uk/initial_setup.py
nikhilwoodruff: It's actually just to have a stored file keeping track of the default dataset

PR #243, opened 2021-09-28
Add safety measures

nikhilwoodruff:
Fixes #242 . Overall, I'm not entirely sure if any of these fixed the issue - these are extra safety measures that I tried independently and didn't work, and the issue was fixed when I regenerated the FRS datasets. But I think all of these are good to add in even just to prevent future bugs.

PR #244, opened 2021-09-28
Safety measures

nikhilwoodruff:
From #243

### DIFF
--- a/openfisca_uk/variables/finance/tax/income_tax/liability.py
+++ b/openfisca_uk/variables/finance/tax/income_tax/liability.py
@@ -488,6 +488,7 @@ class is_higher_earner(Variable):
 
     def formula(person, period, parameters):
         income = person("adjusted_net_income", period)
+        # Add noise to incomes in order to avoid ties
         return (
             person.get_rank(person.benunit, -income + random(person) * 1e-2)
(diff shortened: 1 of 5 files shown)


openfisca_uk/variables/finance/tax/income_tax/liability.py
```diff
@@ -488,7 +488,10 @@ class is_higher_earner(Variable):
 
     def formula(person, period, parameters):
         income = person("adjusted_net_income", period)
-        return person.get_rank(person.benunit, -income) == 0
+        return (
+            person.get_rank(person.benunit, -income + random(person) * 1e-2)
```
MaxGhenis, line 492: could you leave a comment describing this?

MaxGhenis (approved):

openfisca_uk/variables/finance/tax/income_tax/liability.py
nikhilwoodruff: Yep. In case it's not clear here, I was wondering whether if partners who have exactly the same income would mistakenly both be liable for the HITC, which is paid by the higher earner. But this didn't work and I'm pretty sure it's handled somewhere else also.

PR #310, opened 2021-10-31
Add postcode -> country path

nikhilwoodruff:
Any inputs on the path:
* Postcode
* Local Authority, BRMA
* Region
* Country

will propagate upwards, enabling PolicyEngine to provide localised housing costs support figures.

### DIFF
--- a/openfisca_uk/tests/policy/reforms/with_postcode_features/brma.yaml
+++ b/openfisca_uk/tests/policy/reforms/with_postcode_features/brma.yaml
@@ -1,6 +1,0 @@
-- name: BRMA from postcode sector
-  period: 2021
-  input:
-    postcode_sector: RH98
-  output:
-    BRMA: ASHFORD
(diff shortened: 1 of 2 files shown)


openfisca_uk/config/postcode_lookup.py
```diff
@@ -1,51 +1,131 @@
...
+
+class region(Variable):
+    value_type = Enum
+    possible_values = Region
+    default_value = Region.UNKNOWN
+    entity = Household
+    label = u"Region"
+    documentation = "Area of the UK"
+    definition_period = ETERNITY
+    metadata = dict(
```
MaxGhenis, line 109: do we still need this?

MaxGhenis (approved):

nikhilwoodruff:
Closing (but keeping as a reference) because I think this would need a re-implementation: I think we can do this without needing the postcode dataset (which significantly increases the size of the repo).

PR #321, opened 2021-11-08
Add support for extra bands

nikhilwoodruff:
There was a small bug that meant that adding extra bands would break a list mapping.


nikhilwoodruff:
Merging due to small size and to get PolicyEngine PR running

PR #327, opened 2021-11-09
Match regional populations in weights

### LINKED ISSUE #319: Separate population uprating by country
Per this table from https://www.ons.gov.uk/peoplepopulationandcommunity/populationandmigration/populationprojections/bulletins/nationalpopulationprojections/2018based.

image

nikhilwoodruff:
Should fix #319 - @MaxGhenis how does this look? I just need to connect the new weights to the microsim interface, but thought I'd check first.


MaxGhenis (commented):
Thanks - do you have a script for generating the yaml file?

Could you stitch it together with the population projections linked in https://github.com/PolicyEngine/openfisca-uk/issues/319, at least through 2023, to capture population increase from 2020 to 2021?

The yaml also needs some metadata.

nikhilwoodruff:
Dropped the regional populations in favour of country-level figures. This also now requires the automatic parameter interpolation tools in https://github.com/PolicyEngine/openfisca-tools/pull/3 and the changes to the personal weight variable loaded in https://github.com/PolicyEngine/openfisca-uk-data/pull/49.

openfisca_uk/parameters/demographic/population_estimate.yaml
```diff
@@ -0,0 +1,37 @@
...
+      interval: month
+  values:
+    2018-01-01: 56_000_000
+    2023-01-01: 57_600_000
+    2028-01-01: 58_800_000
+    2033-01-01: 59_800_000
+NORTHERN_IRELAND:
+  metadata:
+    interpolation:
+      interval: month
```
MaxGhenis, line 14: Can we make this quarterly to align with the UC rollout?

openfisca_uk/variables/demographic/person.py
```diff
@@ -18,14 +18,39 @@ class people(Variable):
...
+    label = "Weight (region-adjusted)"
+    documentation = "Adjusted to match regional population estimates"
+    definition_period = YEAR
+
+    def formula(person, period, parameters):
+        nation = person.household("country", period)
+        national_population = parameters(
+            period
+        ).demographic.population_estimate[nation]
+        national_weight_sum = (
```
MaxGhenis, line 42: would a `where` statement be faster/more intuitive?

MaxGhenis, line 42: Suggest adding `frs_person_weight = person("FRS_person_weight", period)` to use in the two places it's called.

openfisca_uk/variables/demographic/person.py
```diff
@@ -18,14 +18,39 @@ class people(Variable):
...
+            pd.Series(person("FRS_person_weight", period))
+            .groupby(nation)
+            .sum()[nation]
+        )
+        return (
```
MaxGhenis, line 47:
```suggestion
        scale_factor = national_population / national_weight_sum
        return frs_person_weight * scale_factor
```

MaxGhenis (approved):

openfisca_uk/parameters/demographic/population_estimate.yaml
```diff
@@ -0,0 +1,37 @@
+description: Population estimates for UK nations
+ENGLAND:
+  metadata:
```
MaxGhenis, line 3: I'm guessing we need `metadata` for each nation for it to work with the interpolation. If this is correct could you add a comment noting as such?

openfisca_uk/variables/demographic/person.py
nikhilwoodruff: How would that work?

openfisca_uk/parameters/demographic/population_estimate.yaml
nikhilwoodruff: Actually that'd make it significantly more complex - the current implementation makes use of OpenFisca's periods (day, month, year). We could, but we'd need to add `quarter` as a period.

openfisca_uk/variables/demographic/person.py
MaxGhenis: Ah never mind

openfisca_uk/parameters/demographic/population_estimate.yaml
nikhilwoodruff: Yep, added

nikhilwoodruff:
Not to merge without verifying tests pass across this, the data repo and policyengine.

openfisca_uk/parameters/demographic/population_estimate.yaml
```diff
@@ -0,0 +1,38 @@
+description: Population estimates for UK nations
+# Metadata required for each country parameter
+ENGLAND:
+  metadata:
+    interpolation:
+      interval: month
+  values:
+    2018-01-01: 56_000_000
```
MaxGhenis, line 8: Stitch together actuals and set to mid-year

nikhilwoodruff:
OK, just got the tests passing when pointing to the branch for https://github.com/PolicyEngine/openfisca-uk-data/pull/49. Going to fall back to pip install and merge this after that one.

PR #351, opened 2021-11-15
Param uprating

nikhilwoodruff:
Adds improvements to poverty calculations, including:
* Using primary sources and calculating poverty thresholds from said sources (i.e. from the actual definition - 60% of median income bhc/ahc in 2010, uprating with CPI, rather than taking an intermediate IFS result from 2015 and uprating from that year)
* Using the new parameter uprating tools in openfisca-tools (requires https://github.com/PolicyEngine/openfisca-tools/pull/6)
* Add a short summary-printing script, for easy diff viewing in future (could be executed automatically in future, but the functionality is now there should we want to check future PRs)

### DIFF
--- a/docs/summary/summary.yaml
+++ b/docs/summary/summary.yaml
@@ -1,7 +1,7 @@
-Benefit expenditure: 222bn
-Employment income: 832bn
-Income Tax revenue: 180bn
+Benefit expenditure: 222.4bn
+Employment income: 832.4bn
+Income Tax revenue: 179.6bn
 National Insurance (employee-side) revenue: 64.2bn
 Poverty rate (AHC): 18.78%
 Poverty rate (BHC): 15.95%
-Total income: 1tr
+Total income: 1,222.7bn
--- a/docs/summary/summary.yaml
+++ b/docs/summary/summary.yaml
@@ -1,0 +1,7 @@
+Benefit expenditure: 222bn
+Employment income: 832bn
+Income Tax revenue: 180bn
+National Insurance (employee-side) revenue: 64.2bn
+Poverty rate (AHC): 18.78%
+Poverty rate (BHC): 15.95%
+Total income: 1tr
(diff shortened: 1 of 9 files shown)


docs/summary/summary.yaml
```diff
@@ -0,0 +1,7 @@
+Benefit expenditure: 222bn
+Employment income: 832bn
+Income Tax revenue: 180bn
+National Insurance (employee-side) revenue: 64.2bn
+Poverty rate (AHC): 18.78%
+Poverty rate (BHC): 15.95%
+Total income: 1tr
```
MaxGhenis, line 7: could we add some decimals?

openfisca_uk/parameters/poverty/absolute_poverty_threshold_ahc.yaml
```diff
@@ -1,8 +1,12 @@
 description: Absolute poverty threshold for equivalised household net income, after housing costs.
 values:
-  2015-06-01:
-    value: 270
+  2010-01-01:
```
MaxGhenis, line 3: should this be 2010-04-01, or does it have to align with the CPI dates?

MaxGhenis (approved):
LGTM after running black and the newline script.

This will just uprate yearly right?

openfisca_uk/parameters/poverty/absolute_poverty_threshold_ahc.yaml
nikhilwoodruff: Yeah, the way the uprating is implemented, it won't make a difference unless we interpolate CPI further

docs/summary/summary.yaml
nikhilwoodruff: Opted for straight-forward rounding to 2dp

nikhilwoodruff:
Gah, missed the newline script.

PR #397, opened 2021-12-10
Child benefit

nikhilwoodruff:
Reorganises, adds metadata and references for Child Benefit.


openfisca_uk/parameters/hmrc/benefit/child_benefit/amount/additional.yaml
```diff
@@ -0,0 +1,45 @@
...
+  2015-04-06:
+    value: 13.70
+    reference:
+      - title: The Child Benefit and Tax Credits Up-rating Order 2015(2)(b)
+        href: https://www.legislation.gov.uk/uksi/2015/567/article/2/made
+  2020-04-06:
+    value: 13.95
+    reference:
+      - title: The Tax Credits, Child Benefit and Guardian’s Allowance Up-rating Regulations 2020(5)(b)
+        href: https://www.legislation.gov.uk/uksi/2020/298/contents/made
```
MaxGhenis, line 35: I think should be this:
```suggestion
        href: https://www.legislation.gov.uk/uksi/2020/298/regulation/5/made
```

openfisca_uk/tests/policy/baseline/hmrc/benefit/child_benefit/child_benefit_respective_amount.yaml
```diff
@@ -0,0 +1,41 @@
...
+        age: 12
+      younger:
+        age: 11
+      youngest:
+        age: 10
+    benunits:
+      benunit:
+        children: [eldest, younger, youngest]
+  output:
+    child_benefit_respective_amount: [1099.80, 728.00, 728.00]
```
MaxGhenis, line 41: newline

openfisca_uk/variables/hmrc/benefit/child_benefit.py
```diff
@@ -0,0 +1,77 @@
...
+class child_benefit(Variable):
+    label = "Child Benefit"
+    documentation = "Total Child Benefit for the benefit unit"
+    entity = BenUnit
+    definition_period = YEAR
+    value_type = float
+    unit = "currency-GBP"
+
+    def formula(benunit, period):
+        entitlement = benunit.sum(
```
MaxGhenis, line 58: Seems like there's some extra complication here arising from assigning CB to individual children. Do we need to do that? Seems similar to assigning taxes to household heads. Though I see that the reported CB is at the person level.

If we can just implement CB as it's paid (to benunits), we can remove `child_benefit_respective_amount` and `is_eldest_child` and apply this formula instead:
```suggestion
        eligible_children = benunit.sum(benunit.members("is_child_or_QYP", period))
        amount = parameters(period).hmrc.benefit.child_benefit.amount
        eldest_entitlement = amount.eldest * min_(eligible_children, 1)
        additional_entitlement = amount.additional * max_(eligible_children - 1, 0)
        total_entitlement = eldest_entitlement + additional_entitlement
        return benunit("would_claim_child_benefit", period) * total_entitlement
```

openfisca_uk/parameters/hmrc/benefit/child_benefit/amount/additional.yaml
```diff
@@ -0,0 +1,45 @@
+description: Child Benefit amount for each additional child
```
MaxGhenis, line 1: Should this path be `hmrc/benefit/child_benefit` or `hmrc/child_benefit`?

In openfisca-us I think we're moving to these folders being either agencies or programs. Child Benefit is listed as a top level category in the [HMRC section of gov.uk](https://www.gov.uk/government/organisations/hm-revenue-customs), so I think we could skip `/benefit`. Notably, Tax Credits are also their own section, so I think that layer (rather than WTC and CTC being under HMRC) still makes sense.

MaxGhenis (commented):
Super clean, nice.

openfisca_uk/variables/hmrc/benefit/child_benefit.py
nikhilwoodruff: Yeah I did think about this. But I do favour computing first the child benefit for each child, then adding, because it seems like a good level of decomposition.

openfisca_uk/parameters/hmrc/benefit/child_benefit/amount/additional.yaml
nikhilwoodruff: Moved to hmrc/child_benefit, thanks

PR #408, opened 2021-12-21
Add parameters and re-organise tax charges and reliefs

nikhilwoodruff:
Adds necessary metadata (and some housekeeping) for:
* Child Benefit HITC
* Marriage Allowance (there was actually a bug in this, preventing it from being received - an issue with partner/self income distinction)
* Personal Allowance phase-out rate

PR #410, opened 2021-12-21
Fix child benefit bug

nikhilwoodruff:
In #408 I introduced a bug where high earners would continue to overpay the HITC as their income rises - this fixes it.


MaxGhenis:
Could you add a test for this?

nikhilwoodruff:
Yeah- for context, it got caught by the PolicyEngine tests, but not OpenFisca-UK

PR #433, opened 2021-12-27
Add tax credits validation page

nikhilwoodruff:
@MaxGhenis  this makes the UKMOD test redundant, so we can remove it.


MaxGhenis:
Nice, the [rendered notebook](https://nbviewer.org/github/PolicyEngine/openfisca-uk/blob/3dd9137c50bd10890f8ac9a5ea24aeae7f052ecc/docs/book/model/tax_credits.ipynb) looks good. Couple minor suggestions:
* Change units from £m to £bn
* Change series order to [official, openfisca, ukmod]
* Change colors to {official: black, openfisca: blue, ukmod: gray}

I like the idea of testing that we're at least as close to official as UKMOD is, that's better than comparing against UKMOD. But could we put these sorts of assert statements in a Python file so we get informative failure messages in the GitHub Actions?

MaxGhenis:
Should we add `docs/*` to this? I could see the case for either including documentation updates in the changelog and not, though seems odd to bump the Python package version.

https://github.com/PolicyEngine/openfisca-uk/blob/24aed7b9daa96c3a2c1028d9e4de125e8cdc83c4/.github/has-functional-changes.sh#L3

openfisca_uk/tests/microsimulation/test_statistics.py
```diff
@@ -19,34 +19,104 @@ def get_openfisca_uk_aggregate(variable: str, year: int) -> float:
...
+        return openfisca_uk_error < self.max_error
+
+    def describe(self):
+        return f"OpenFisca-UK {self.variable_label} {self.statistic} absolute error is less than {self.max_error} in {self.year}"
+
+
+class RelativeErrorLessThan(StatisticTest):
+    def test(self):
+        openfisca_uk_error = abs(self.openfisca_uk - self.official)
+        return openfisca_uk_error / self.official < self.max_error
```
MaxGhenis:
```suggestion
        relative_error = absolute_error / self.official
        return relative_error < self.max_error
```

openfisca_uk/tests/microsimulation/test_statistics.py
```diff
@@ -19,34 +19,104 @@ def get_openfisca_uk_aggregate(variable: str, year: int) -> float:
...
+        openfisca_uk_error = abs(self.openfisca_uk - self.official)
+        return openfisca_uk_error < self.max_error
+
+    def describe(self):
+        return f"OpenFisca-UK {self.variable_label} {self.statistic} absolute error is less than {self.max_error} in {self.year}"
+
+
+class RelativeErrorLessThan(StatisticTest):
+    def test(self):
+        openfisca_uk_error = abs(self.openfisca_uk - self.official)
```
MaxGhenis:
```suggestion
        absolute_error = abs(self.openfisca_uk - self.official)
```

openfisca_uk/variables/finance/expense.py
```diff
@@ -169,7 +169,10 @@ class council_tax_less_benefit(Variable):
     def formula(household, period, parameters):
         return (
```
MaxGhenis:
```suggestion
        person = household.members
        council_tax_benefit = household.sum(person.benunit("council_tax_benefit", period) * person("is_benunit_head"))
        return household("council_tax") - council_tax_benefit
```

nikhilwoodruff:
> Should we add `docs/*` to this? I could see the case for either including documentation updates in the changelog and not, though seems odd to bump the Python package version.
> 
> https://github.com/PolicyEngine/openfisca-uk/blob/24aed7b9daa96c3a2c1028d9e4de125e8cdc83c4/.github/has-functional-changes.sh#L3

Yeah OK so this definitely changes the version, but I've added it like you suggest

Issue #418, opened 2021-12-24
Replace `num_*` with `count_*` for variables that indicate number of something


MaxGhenis:
`num_` can be confused for an identifier, the online [consensus](https://stackoverflow.com/a/21345217/1840471) suggests that `count` is clearer.

Example:

https://github.com/PolicyEngine/openfisca-uk/blob/1f788df0c8307fc9e6b93c15d4e67f53da0581b0/openfisca_uk/variables/demographic/household.py#L164

vahid-ahmadi:
Audit before anyone takes this on:

**State**: 11 `num_*` variables still exist:
- `num_children`, `num_adults`, `num_carers`, `num_bedrooms`, `num_vehicles`
- `num_disabled_adults`, `num_disabled_children`
- `num_severely_disabled_adults`, `num_severely_disabled_children`
- `num_enhanced_disabled_adults`, `num_enhanced_disabled_children`

Referenced in 27+ places inside `policyengine-uk` and at minimum in [`policyengine-uk-data`](https://github.com/PolicyEngine/policyengine-uk-data) (`docs/imputations.md`, `targets/sources/dwp.py`, `datasets/imputations/wealth.py`, `datasets/imputations/consumption.py`) — likely also in `policyengine-api`, `policyengine-app` / `-v2`, and any external scripts hitting the household API.

**Trade-off**:
- *Benefit*: marginal readability gain (`num_children` → `count_children` or `children`). The qualifier is unambiguous in context — these are all numeric counts of household members/items.
- *Cost*: breaking rename across 4-5 repos and any external API consumer. Backwards-compat aliases would mean carrying both names forever, which is worse than the original concern.

If we want to do this, it should be planned as a single coordinated cross-repo change with a deprecation window — not a piecemeal rename in this repo. Leaving open as a discussion anchor; happy to implement once there's a green light and an agreed deprecation strategy.

PR #441, opened 2022-01-06
Add parameter for CTC child limit, rename UC child limit parameter

### LINKED ISSUE #439: Parameterize two-child CTC limit
Mirror [`benefit.universal_credit.elements.child.limit.num_children`](https://github.com/PolicyEngine/openfisca-uk/blob/master/openfisca_uk/parameters/benefit/universal_credit/elements/child/limit/num_children.yaml)

Currently, it's hard-coded via [`benefit.tax_credits.child_tax_credit.two_child_limit_year`](https://github.com/PolicyEngine/openfisca-uk/blob/master/openfisca_uk/parameters/benefit/tax_credits/child_tax_credit/two_child_limit_year.yaml) and

https://github.com/PolicyEngine/openfisca-uk/blob/24990eca889332d89856c72f7e5bf9881fbc211b/openfisca_uk/variables/finance/benefit/family/tax_credits.py#L166

MaxGhenis:
Fixes #439

- Add a new feature (for instance adding a variable)


openfisca_uk/variables/finance/benefit/family/tax_credits.py
```diff
@@ -50,16 +50,14 @@ def formula(benunit, period, parameters):
...
+    label = u"Exemption from Child Tax Credit child limit"
+    documentation = "Exemption from Child Tax Credit limit on number of children based on birth year"
     definition_period = YEAR
 
     def formula(person, period, parameters):
         limit_year = parameters(
             period
-        ).benefit.tax_credits.child_tax_credit.two_child_limit_year
+        ).benefit.tax_credits.child_tax_credit.limit.start_year
         return person("birth_year", period.this_year) <= limit_year
```
MaxGhenis: I didn't get a chance to add legislative references here, but this differs from the equivalent UC code:
```python
born_before_limit = person("birth_year", period) < start_year
```
`period` vs `period.this_year` and `<` vs `<=`

openfisca_uk/variables/finance/benefit/family/universal_credit.py
```diff
@@ -153,7 +153,7 @@ def formula(person, period, parameters):
         child_index = person("child_index", period)
         born_before_limit = person("is_child_born_before_child_limit", period)
         child_limit_applying = where(
-            ~born_before_limit, UC.elements.child.limit.num_children, 1e2
+            ~born_before_limit, UC.elements.child.limit.child_count, 1e2
```
MaxGhenis: Changed per #418 

I tried just `children`, but hit https://github.com/openfisca/openfisca-core/issues/1101

MaxGhenis: Change 1e2 to inf

openfisca_uk/variables/finance/benefit/family/tax_credits.py
MaxGhenis: @nikhilwoodruff this is the only lingering question in this PR

nikhilwoodruff: OK - since it's yearly, `period === period.this_year` (so fine either way IMO, though `period` would be shorter). On the `<` issue - we are already simplifying from the law: in reality, the birth date needs to be before April 2017. So if it's just a question of which operation is closer to the law, `<` is better.

MaxGhenis: OK updated to match UC. This made us further from the administrative total though, so I had to increase the tolerance from 7bn to 8bn to pass tests.

PR #474, opened 2022-01-17
Stocks and flows

nikhilwoodruff:
### Added

* Stocks/flows metadata for PolicyEngine-facing variables.


MaxGhenis:
Is either stock or flow a default?

PR #505, opened 2022-01-30
Re-weight FRS households

### LINKED ISSUE #504: Adjust FRS weights to match administrative statistics
The FRS has a well-documented problem with benefit reporting and high incomes: all benefits are under-reported, as well as income sources such as dividends. Up to now, we've operated under the assumption that this is due to measurement error: that recipients are giving incorrect information about their income. However, over time I think we've discovered enough evidence to suggest it is actually mostly sampling error:
* When we calibrate takeup rates to match administrative totals, we are unable to match both caseload and expenditure statistics at the same time, suggesting that the FRS recipients are not highly representative of actual recipients. See #495.
* The [FRS questionnaire](https://sp.ukdataservice.ac.uk/doc/6523/mrdoc/pdf/frs_2008_09_gb_question_instructions.pdf) is highly detailed, with multiple checks on users' answers, and surveyors ask for documented proof of incomes. This still leaves room for error if individuals are not compliant or not aware of benefit receipt, but I doubt to the extent that the data under-reports reality (around 30%).
* UKMOD, the only other microsimulation model that publishes validation statistics, [also underestimates benefit aggregates and tax revenues](https://www.iser.essex.ac.uk/research/publications/working-papers/cempa/cempa7-20.pdf#127). Under-estimation can be caused by either incorrect data or incorrect modelling: UKMOD and OpenFisca-UK have separate implementations of the same programs, suggesting modelling is not the issue.

After some [initial trials](https://colab.research.google.com/drive/1fOKdvH4eEiM8fNXAc6KJkjW9EIZpY3oJ#scrollTo=K73JA_sycWWk), I propose an optimisation-based approach at re-weighting, using TensorFlow to optimise a weight adjustment vector in order to minimise statistical error across a range of validation statistics, penalising substantial divergence from initial weights. Essentially, we want a balance between modified weights that move us as close as possible to benefit and tax statistics, while not moving too far from the initial weights.

### Initial experimentation

The following graph illustrates the trade-off between weight edits and statistical error:
image

And here's an example result from one specific benefit and metric. We can get closer or further, depending on the modification penalty:
image

### Process outline

We'll aim to match the following targets:
* Caseloads and aggregates for all eight major means-tested benefits, plus Income Tax and National Insurance.
* UK and regional household populations

cc @MaxGhenis

nikhilwoodruff:
Fixes #504


openfisca_uk/parameters/statistics.yaml
```diff
@@ -0,0 +1,253 @@
+aggregate:
```
MaxGhenis: Aggregate could refer to either expenditure or participation.

Could you also confirm that these are all expenditure received by households, and don't include administrative overhead?
```suggestion
expenditure:
```

openfisca_uk/parameters/statistics.yaml
```diff
@@ -0,0 +1,253 @@
...
+      2018-01-01: 192_500_000_000
+      2019-01-01: 193_600_000_000
+      2020-01-01: 188_200_000_000
+      2021-01-01: 203_600_000_000
+      2022-01-01: 205_000_000_000
+    metadata:
+      reference:
+        - title: OBR 2021 forecast
+          href: https://obr.uk/forecasts-in-depth/tax-by-tax-spend-by-spend/income-tax/
+count:
```
MaxGhenis:
```suggestion
participation:
```

openfisca_uk/parameters/statistics.yaml
```diff
@@ -0,0 +1,253 @@
...
+
+  child_tax_credit:
+    description: Number of families receiving Child Tax Credit
+    values:
+      2018-01-01: 3_434_000
+      2019-01-01: 2_835_000
+      2020-01-01: 2_125_000
+      2021-01-01: 1_686_000
+    metadata:
+      unit: currency-GBP
```
MaxGhenis:
```suggestion
      unit: benefit unit
```

openfisca_uk/parameters/statistics.yaml
```diff
@@ -0,0 +1,253 @@
...
+
+  income_tax:
+    description: Number of individuals paying Income Tax.
+    values:
+      2018-01-01: 31_600_000
+      2019-01-01: 31_400_000
+      2020-01-01: 31_700_000
+      2021-01-01: 38_272_000
+    metadata:
+      reference:
```
MaxGhenis:
```suggestion
      unit: person
      reference:
```

openfisca_uk/parameters/statistics.yaml
```diff
@@ -0,0 +1,253 @@
...
+
+  working_tax_credit:
+    description: Number of families receiving Working Tax Credit
+    values:
+      2018-01-01: 1_970_000
+      2019-01-01: 1_615_000
+      2020-01-01: 1_199_000
+      2021-01-01: 1_044_000
+    metadata:
+      unit: currency-GBP
```
MaxGhenis:
```suggestion
      unit: benefit unit
```

openfisca_uk/parameters/statistics.yaml
```diff
@@ -0,0 +1,253 @@
...
+          href: https://www.gov.uk/government/publications/benefit-expenditure-and-caseload-tables-2021
+
+  child_benefit:
+    description: Number of families claiming Child Benefit.
+    values:
+      2018-01-01: 7_325_845
+      2019-01-01: 7_281_535
+      2020-01-01: 7_210_595
+    metadata:
+      reference:
```
MaxGhenis:
```suggestion
      unit: benefit unit
      reference:
```

openfisca_uk/parameters/statistics.yaml
nikhilwoodruff: OK - though there are tax revenues later down, the name isn't super important though.

On the admin inclusion, I don't think these include administrative costs. It's not 100% clear because the benefit and caseload tables don't specify admin expenditure, but I did find some data on admin costs from the [2016-17 DWP accounts](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/624319/dwp-annual-report-and-accounts-2016-2017.pdf#page=122), and the 2015-16 outturn data (which has zero administrative costs, as they're specified elsewhere) for individual benefits in those tables matches with the benefits and caseload tables figures. I think this means the admin costs are not included in these figures.

openfisca_uk/calibration/reweighting.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+        for variable in variables:
+            if variable in statistic_set.aggregate._children:
+                values = sim.calc(
+                    variable, period=year, map_to="household"
+                ).values
+                # Calculate aggregate error
+                agg = tf.reduce_sum(modified_weights * values)
+                target = getattr(statistic_set.aggregate, variable)
+                l += ((agg - target) / 1e9) ** 2
```
MaxGhenis: Could you move this 1e9 into a named constant like `MODIFICATION_PENALTY` (also the 1e6 for caseload)

Alternatively, we could normalize each deviation by dividing by `target` and then specify an importance vector for each variable (possibly set at the current aggregates with a 1000x boost for participation to mimic the current logic).

openfisca_uk/calibration/reweighting.py
```diff
@@ -0,0 +1,358 @@
...
+household_region = sim.calc("region").values
+regions = list(pd.Series(household_region).unique())
+household_country = sim.calc("country").values
+countries = list(pd.Series(household_country).unique())
+survey_num_households = len(sim.calc("household_id"))
+
+AGGREGATE_ERROR_PENALTY = 1e-5
+PARTICIPATION_ERROR_PENALTY = 1e-1  # Person-deviations are 10,000 more loss-heavy than spending deviations per pound
+MODIFICATION_PENALTY = 1e-11
+AGE_DISTRIBUTION_PENALTY = 1e-1
```
MaxGhenis: Looks like this is to shrink each age group's population penalty by 10, if that's right could you leave a comment?

openfisca_uk/calibration/reweighting.py
```diff
@@ -0,0 +1,358 @@
...
+                household_participants = sim.map_to(
+                    (values > 0) * 1, entity, "household"
+                )
+                # Calculate aggregate error
+                agg = tf.reduce_sum(modified_weights * household_totals)
+                target = getattr(statistic_set.aggregate, variable)
+                l += (
+                    AGGREGATE_ERROR_PENALTY
+                    * target
+                    * ((agg / target) - 1) ** 2
```
MaxGhenis: Could you make a function for this e.g. `squared_relative_deviation(value, target)`? Since it's used below too. Might be room for modularizing other stuff too, but this would at least be a bit more descriptive.

openfisca_uk/calibration/reweighting.py
```diff
@@ -0,0 +1,358 @@
...
+                target = getattr(
+                    getattr(statistic_set.aggregate_by_country, variable),
+                    country,
+                )
+                l += (
+                    AGGREGATE_ERROR_PENALTY
+                    * target
+                    * ((aggregate / target) - 1) ** 2
+                )
+        # Income tax payers by tax band
```
MaxGhenis: Yeah actually this section could be its own function to break up the code

openfisca_uk/calibration/reweighting.py
```diff
@@ -0,0 +1,358 @@
...
+
+        l += (
+            POPULATION_ERROR_PENALTY
+            * PARTICIPATION_ERROR_PENALTY
+            * weights.sum()
+            * ((tf.reduce_sum(modified_weights) / weights.sum()) - 1) ** 2
+        )
+
+    # Add penalty for weight changes
+    if include_modification_penalty:
```
MaxGhenis: Why the `if` statement instead of setting the penalty to zero?

nikhilwoodruff: Code-character-count efficiency? No real preference

openfisca_uk/calibration/reweighting.py
nikhilwoodruff: Yeah plus the others, done

nikhilwoodruff:
I've also just removed the UKMOD tests here. I think we can do this - aggregates and caseloads are now all tested against external statistics rather than UKMOD aggregates. There might still be a case for a limited number of variables to check policy logic for specific inputs, but we're already unit testing and many have different structures.

nikhilwoodruff:
JB seems to be failing here but succeeding locally, checking.

MaxGhenis:
Update: We discovered that this result significantly improves fit on the included metrics, but worsens fit on State Pension + Pension Credit repeal, indicating overfitting.

Next step: Creating a "validation" metric, and running the routine on different levels of the weight modification penalty to optimize the penalty against out-of-sample metrics.

nikhilwoodruff:
Fixes #507

nikhilwoodruff:
@MaxGhenis FYI [this change](https://github.com/PolicyEngine/openfisca-uk/pull/505/commits/f6de27d657b2cf34375ba29b0fb4147648443cc0) was an error - because the loss function `UK population` was misnamed (it actually tests the number of households, rather than number of people), I thought it was covered by the regional population function. Re-added and re-named.

MaxGhenis (approved):

PR #556, opened 2022-03-24
Re-calibrate with forecasting

nikhilwoodruff:
Updates the calibration makefile, ensuring we use the same number of epochs (200).


nikhilwoodruff:
Superseded by #558

PR #562, opened 2022-04-05
Fix NI threshold

nikhilwoodruff:


nikhilwoodruff:
Actually, after reading the [HMRC explanatory notes](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1062667/Explanatory_notes_National_Insurance_contributions__increase_of_thresholds__Bill.pdf) and the [draft bill](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1062566/Draft_legislation_National_Insurance_Contributions__Increase_of_Thresholds__Bill.pdf), I think it's clear that the self-employed Class 4 thresholds come into effect on 6th April, rather than in July.

PR #574, opened 2022-04-14
Re-add MANIFEST.in

nikhilwoodruff:
I think this was the root cause of the issues the previous two PRs aimed to solve.


nikhilwoodruff:
The reason this issue was a bit hard to find was because the MANIFEST.in file only caused issues when the package is installed from PyPI, and not when installed from a clone locally (which I'm doing when debugging).

PR #577, opened 2022-04-15
Household basic income phase-outs

### LINKED ISSUE #576: Phase out full family's UBI
Currently, the UBI phase-out only applies to the primary household member's UBI. For example, here's how a [£200/week UBI for all age groups phasing out at 50%](https://policyengine.org/uk/household?adult_bi=200&child_bi=200&senior_bi=200&bi_phase_out_rate=50) affects a married family with two children:

image

The MTR change stops at £20,800, which is double the primary earner's UBI (due to 50% phase-out):
image

This occurs because the [`bi_phaseout` variable](https://github.com/PolicyEngine/openfisca-uk/blob/master/openfisca_uk/variables/contrib/ubi_center/basic_income/bi_phaseout.py) is at the person level.

nikhilwoodruff:
Fixes #576


openfisca_uk/variables/contrib/ubi_center/basic_income/bi_phaseout.py
```diff
@@ -12,6 +15,56 @@ def formula(person, period, parameters):
...
+    value_type = float
+    unit = "currency-GBP"
+
+    def formula(person, period, parameters):
+        income = person("total_income", period)
+        household = person.household
+        household_income = household.sum(income)
+        bi = parameters(period).contrib.ubi_center.basic_income
+        remaining_bi = person("bi_maximum", period) - person(
+            "bi_individual_phaseout", period
```
nikhilwoodruff, line 43: Split out

openfisca_uk/tests/policy/reforms/parametric/basic_income/phase_out.yaml
```diff
@@ -0,0 +1,23 @@
+- name: Individual and household phase-outs together
+  period: 2022
+  absolute_error_margin: 1
+  input:
+    contrib.ubi_center.basic_income.phase_out.individual.rate: 1
```
nikhilwoodruff, line 5: Add simpler unit test

PR #597, opened 2022-04-27
Begin refactoring and adding statistics

nikhilwoodruff:
* Adds country-level statistics for benefits


openfisca_uk/parameters/calibration/programs/income_tax/revenue/by_country.yaml
```diff
@@ -0,0 +1,20 @@
...
+metadata:
+  uprating: calibration.programs.income_tax.revenue.in_total
+  propagate_metadata_to_children: true
+  reference:
+    - title: HMRC SPI Analysis
+      href: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/853118/Disaggregated_tax_and_NICs_receipts_-_methodological_note.pdf#page=14
+ENGLAND:
+  values:
+    2018-01-01: 169_958_000_000
+  metadata:
```
MaxGhenis: intended?

openfisca_uk/parameters/calibration/programs/total_NI/revenue/in_total.yaml
```diff
@@ -0,0 +1,15 @@
+total_NI:
```
MaxGhenis: could you spell out as `national_insurance`?

nikhilwoodruff: Oh here the names are the same as the OpenFisca-UK variable names: `national_insurance` is just employee-side, whereas `total_NI` includes employer contributions. Should we rename them?

openfisca_uk/parameters/calibration/programs/income_tax/revenue/by_country.yaml
nikhilwoodruff: Ah good catch

openfisca_uk/parameters/calibration/programs/total_NI/revenue/in_total.yaml
MaxGhenis: Ah, yes I think we should but doesn't have to be here

openfisca_uk/data/datasets/frs/enhanced/stages/calibration/losses/loss_category.py
```diff
@@ -21,7 +21,7 @@ def weighted_squared_relative_deviation(
     Returns:
         tf.Tensor: The weighted squared relative deviation.
     """
-    return (pred - actual) ** 2
+    return (pred / (actual + 1) - 1) ** 2
```
MaxGhenis, line 25: Is this because there are actuals of zero? What's an example if so?

MaxGhenis (approved):

openfisca_uk/data/datasets/frs/enhanced/stages/calibration/losses/loss_category.py
nikhilwoodruff: Yes, legacy benefits which are phased out. Also the I Council Tax band for some regions.

PR #605, opened 2022-05-22
Generate takeup-informing variables before calibration, not after

nikhilwoodruff:
I think this is what's causing some overestimation of Universal Credit participants (~600k families). Fixed when re-running the dataset generation after these changes are applied.


openfisca_uk/variables/gov/hmrc/child_benefit.py
```diff
@@ -67,6 +68,16 @@ def formula(person, period, parameters):
         return eligible * amount * WEEKS_IN_YEAR
 
 
+class child_benefit_entitlement(Variable):
+    label = "CB entitlement"
+    entity = BenUnit
+    definition_period = YEAR
+    value_type = float
+    unit = "currency-GBP"
```
MaxGhenis:
```suggestion
    unit = GBP
```

MaxGhenis (approved):
Makes sense, and I like the `entitlement/would_claim` pattern.

PR #638, opened 2022-09-27
Add flat per-person UBI policy parameter

### LINKED ISSUE #637: Add single UBI policy parameter
Add the age-specific amounts to this

MaxGhenis:
Fixes #637

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ format:
 	black . -l 79
 
 test:
-	openfisca test -c openfisca_uk openfisca_uk/tests/policy/baseline
+	openfisca-uk test openfisca_uk/tests/policy/baseline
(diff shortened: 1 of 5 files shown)


MaxGhenis:
@nikhilwoodruff the test works via `openfisca-uk test openfisca_uk/tests/policy/baseline/contrib` but not `make test` (and not on GitHub). Why does `make test` run `openfisca test -c openfisca_uk` instead of `openfisca-uk test`?

nikhilwoodruff:
@MaxGhenis - it's because the `openfisca-uk` CLI has the modifications needed to run reform tests, which makes it slightly slower, so I defaulted to using the original OpenFisca CLI for the baseline tests. Could we move the test to the reform folder?

PR #741, opened 2023-07-17
2 child limit bug fix

### LINKED ISSUE #740: Adjust for two-child 2017 start date in FRS years
<del>The child indexing method appears to be not applying to the 4th child onwards (affecting 1% of households)</del>

Actually, a misdiagnosis. It turns out that the FRS data isn't correctly being adjusted for the fact that the child limit applies only to children born after 2017. Given that we can't just increase the ages of children across survey years (we'd end up with no 6 year olds or younger in 2023), we should adjust the 2017 year parameter downwards within the formula itself for FRS microsimulations as a partial fix.

nikhilwoodruff:
Fixes #740


policyengine_uk/variables/gov/dwp/universal_credit.py
```diff
@@ -182,13 +190,11 @@ def formula(person, period, parameters):
             select(
```
MaxGhenis, line 190: suggestion
```python
amount_if_eligible = where((child_index == 1) & born_before_limit, UC.elements.child.first.higher_amount, UC.elements.child.amount)
return amount_if_eligible * eligible
```

PR #767, opened 2023-10-08
Unpin numpy

### LINKED ISSUE #766: Unpin numpy
The numpy pin should inherit from policyengine-core

MaxGhenis:
Fixes #766


MaxGhenis:
@nikhilwoodruff could you ptal at the test failures?

PR #779, opened 2023-11-13
NI validation

nikhilwoodruff:
* Validates against UKMOD (mostly matching closely, except from structural differences like month/year mid-year policy change averaging).
* Standardises variable names and files, and adds full unit test coverage.


MaxGhenis:
tests failing

PR #782, opened 2023-11-22
Add increase switch for benefit actual amounts

nikhilwoodruff:
cc @MaxGhenis


nikhilwoodruff:
Merging without a test because of the issue with custom reforms and YAML tests, but looks fine on the web app locally:

image

Issue #786, opened 2023-11-23
Map BRMA to region


MaxGhenis:
So we can assign BRMAs within region for LHA

MaxGhenis:
[Thanks, ChatGPT!](https://chat.openai.com/share/c5a4a089-4276-46ec-ae7c-6b92d869e84b) I spot-checked a few and they all looked right.
```yaml
BRMA_to_Region:
  Ashford: "South East"
  Aylesbury: "South East"
  Barnsley: "Yorkshire and the Humber"
  Barrow_in_Furness: "North West"
  Basingstoke: "South East"
  Bath: "South West"
  Bedford: "East of England"
  Birmingham: "West Midlands"
  Black_Country: "West Midlands"
  Blackwater_Valley: "South East"
  Bolton_and_Bury: "North West"
  Bournemouth: "South West"
  Bradford_South_Dales: "Yorkshire and the Humber"
  Brighton_and_Hove: "South East"
  Bristol: "South West"
  Bury_St_Edmunds: "East of England"
  Cambridge: "East of England"
  Canterbury: "South East"
  Central_Greater_Manchester: "North West"
  Central_Lancs: "North West"
  Central_London: "London"
  Central_Norfolk_Norwich: "East of England"
  Chelmsford: "East of England"
  Cheltenham: "South West"
  Cherwell_Valley: "South East"
  Chesterfield: "East Midlands"
  Chichester: "South East"
  Chilterns: "South East"
  Colchester: "East of England"
  Coventry: "West Midlands"
  Crawley_Reigate: "South East"
  Darlington: "North East"
  Derby: "East Midlands"
  Doncaster: "Yorkshire and the Humber"
  Dover_Shepway: "South East"
  Durham: "North East"
  East_Cheshire: "North West"
  East_Lancs: "North West"
  East_Thames_Valley: "South East"
  Eastbourne: "South East"
  Eastern_Staffordshire: "West Midlands"
  Exeter: "South West"
  Fylde_Coast: "North West"
  Gloucester: "South West"
  Grantham_Newark: "East Midlands"
  Greater_Liverpool: "North West"
  Grimsby: "Yorkshire and the Humber"
  Guildford: "South East"
  Halifax: "Yorkshire and the Humber"
  Harlow_Stortford: "East of England"
  Harrogate: "Yorkshire and the Humber"
  Herefordshire: "West Midlands"
  High_Weald: "South East"
[...]

PR #787, opened 2023-11-23
Add Local Housing Allowance parameters

nikhilwoodruff:


policyengine_uk/data/gov/brma_to_region.csv.gz
MaxGhenis, line 1: can you drop a README.md here showing how this was generated? (i'm guessing from the chatgpt mapping in #786)

also the source on the rents data

policyengine_uk/tools/generate_brmas.py
MaxGhenis, line 1: this produces `data/gov/enhanced_frs_brmas.csv.gz`?

MaxGhenis (changes requested):

policyengine_uk/tools/generate_brmas.py
nikhilwoodruff: Yes

PR #799, opened 2023-12-16
Bump policyengine-core

### LINKED ISSUE #798: Bump policyengine-core to 2.11.3
To capture randomness bug fixes

MaxGhenis:
Fixes #798

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+      - Bump policyengine-core to capture simulation randomness bug fixes.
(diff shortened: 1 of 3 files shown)


MaxGhenis:
@nikhilwoodruff can you look into why bumping core would affect these test results?

nikhilwoodruff:
Ah I think the random seed changing would do that

nikhilwoodruff:
Removed MTRs from those tests since I don't think it made sense anyway (we're testing small employment income increases from zero when self-employment is the income source)

nikhilwoodruff (approved):

nikhilwoodruff:
Passing non-microdata tests so merging.

PR #839, opened 2024-03-20
Care to learn

Kittyaoooo:
add care to learn variables


policyengine_uk/parameters/gov/dfe/care_to_learn.py
```diff
@@ -1,22 +1,18 @@
...
-        # if it's £180 per week, the monthly value is just 180 * 52 / 12
-        live_in_london = household("live_in_london", period)
-        # live_in_london = household("region", period).decode_to_str() == "LONDON"
-        amount_per_child = parameters(period).gov.dwp.universal_credit.elements.childcare.childcare_amount_per_child
-        # amount_per_child = parameters(period).gov.dfe.care_to_learn.amount_per_child
+        live_in_london = household("region", period).decode_to_str() == "LONDON"
+        amount_per_child = parameters(period).gov.dfe.care_to_learn.amount_per_child
         if live_in_london:
-            return amount_per_child.IN_LONDON # multiply to get monthly amount
+            return amount_per_child.IN_LONDON
```
nikhilwoodruff, line 16: I think this still returns the weekly amount, no?

nikhilwoodruff (changes requested):

Kittyaoooo:
I've multiplied the data in the amount_per_child.yaml to get monthly
amount, so the value returned should be monthly amount.

nikhilwoodruff:
@Kittyaoooo- sorry, we do need to multiply within the formula. This is because when we add references, they need to match the value in the parameter file, and if we've manually changed it to be monthly, it won't match.

policyengine_uk/parameters/gov/dfe/care_to_learn.py
nikhilwoodruff, line 1: This python file should be with the others in the variables/ folder.

policyengine_uk/tests/policy/baseline/gov/hmrc/income_tax/allowances/allowance_integration.yaml
nikhilwoodruff, line 1: These files are from an earlier PR- could you remove them? I'm not sure why they're here.

PR #846, opened 2024-04-19
Capital Gains Tax improvements

### LINKED ISSUE #833: Subtract Personal Allowance from adjusted net income in CGT
[This code](https://github.com/PolicyEngine/policyengine-uk/blob/7ff3d6e0f918b3f7afe9ff9f1e0eef76b26e1bfc/policyengine_uk/variables/gov/hmrc/capital_gains_tax/capital_gains_tax.py#L19) should be amended as well as adding the AEA.

nikhilwoodruff:
Fixes PolicyEngine/policyengine-uk#833


policyengine_uk/data/datasets/frs/imputations/capital_gains.py
```diff
@@ -18,61 +20,41 @@
...
+            selection = ti.between(lower, upper)
+            first_half_of_selection = selection * first_half
+            second_half_of_selection = selection * ~first_half
+            fraction_with_capital_gains = capital_gains.percent_with_gains.iloc[i]
+            person_weights_subset = person_weight[selection][:selection.sum()//2]
+            person_weight[first_half_of_selection] = person_weights_subset * (1 - fraction_with_capital_gains)
+            person_weight[second_half_of_selection] = person_weights_subset * fraction_with_capital_gains
+            has_capital_gains[second_half_of_selection] = True
+            row = capital_gains.iloc[i]
+            spline = UnivariateSpline(
```
MaxGhenis: To ensure monotonicity, see https://github.com/PolicyEngine/policyengine-uk/issues/816#issuecomment-1962969566
```suggestion
            spline = PchipInterpolator(
```

MaxGhenis (changes requested):

policyengine_uk/data/datasets/frs/imputations/capital_gains.py
nikhilwoodruff: OK so I just tried this and got pretty strange results- capital gains at the high end fell a lot for some reason. We should probably still add it and I'm probably just using it wrong, but needs more careful investigation.

Issue #195, opened 2021-09-03
Add capital gains policy parameters


MaxGhenis:

nikhilwoodruff:
I think the rates from [gov.uk](https://www.gov.uk/capital-gains-tax/rates) could go in a new file at `parameters/tax/capital_gains_tax/rates.yaml`. It's also worth checking legislation.gov.uk for any historical changes.

PR #848, opened 2024-05-01
Added capital gains policy parameters

riyasachdeva04:
Fix for issue #195


nikhilwoodruff:
Thanks for the contribution and sorry for the late review, but I'm going to close this since we already have the parameters modellable, and this addition doesn't fit the pattern of the other parameters.

PR #855, opened 2024-05-10
Add Care-to-learn

Kittyaoooo:
1. add reference to 'amount_per_child.yaml'
2. modify the formula of care_to_learn.py(*MONTHS_IN_YEAR/WEEKS_IN_YEAR)


policyengine_uk/variables/gov/dfe/care_to_learn.py
```diff
@@ -0,0 +1,18 @@
...
+    definition_period = MONTH
+    value_type = float
+    unit = GBP
+    reference = "https://www.gov.uk/care-to-learn/what-youll-get"
+
+    def formula(household, period, parameters):
+        live_in_london = household("region", period).decode_to_str() == "LONDON"
+        amount_per_child = parameters(period).gov.dfe.care_to_learn.amount_per_child
+        if live_in_london:
+            return amount_per_child.IN_LONDON*4
```
nikhilwoodruff: Thanks for making this multiplied! But could we use WEEKS_IN_YEAR and MONTHS_IN_YEAR to get the actual precise multiplier?

Also: this needs to be vectorised. See other examples in the codebase, but a solution would be something like:

```python
return where(live_in_london, amount_per_child.IN_LONDON * MONTHS_IN_YEAR / WEEKS_IN_YEAR, amount...)
```

policyengine_uk/parameters/gov/dfe/care_to_learn/amount_per_child.yaml
```diff
@@ -0,0 +1,18 @@
+metadata:
+  label: care-to-learn subsidy
+
+IN_LONDON:
+  description: Description
+  values:
+    2015-01-01: 195
+  metadata:
+    unit: currency-GBP
+    label: amount per child per month if you live in London
```
nikhilwoodruff, line 12: Thanks- could you add a reference?

nikhilwoodruff (approved):

policyengine_uk/parameters/gov/dfe/care_to_learn/amount_per_child.yaml
nikhilwoodruff, line 12: [Here's an example](https://github.com/PolicyEngine/policyengine-uk/blob/ce8bca6759cb68b8296e7c053ebd29840305b5e9/policyengine_uk/parameters/gov/hmrc/income_tax/rates/uk.yaml#L16)

nikhilwoodruff:
To run the code formatter, run:

`make format`
or
`black . -l 79`

from the main project folder root.

policyengine_uk/variables/gov/dfe/care_to_learn.py
```diff
@@ -0,0 +1,25 @@
...
+    reference = "https://www.gov.uk/care-to-learn/what-youll-get"
+
+    def formula(household, period, parameters):
+        live_in_london = (
+            household("region", period).decode_to_str() == "LONDON"
+        )
+        amount_per_child = parameters(
+            period
+        ).gov.dfe.care_to_learn.amount_per_child
+        multiplier = MONTHS_IN_YEAR / WEEKS_IN_YEAR
```
nikhilwoodruff: This is actually the incorrect multiplier to use- e.g. if the weekly amount is £1, and we want the monthly value, this will be doing 1 * 12 / 52 = 23p, when it actually should be £4.30.

Kittyaoooo: please see the latest version, thank you!

nikhilwoodruff:
Closing due to inactivity!

PR #869, opened 2024-06-07
Reform option to phase out the CB HITC with household income

### LINKED ISSUE #868: Reform option to phase out the CB HITC with household income
As the Tories have [proposed](https://taxpolicy.org.uk/2024/06/06/tories_accidental_70_percent/):

>And to end the unfairness that means single earner households can start paying the tax charge when a household with two working parents and a much higher total income can keep Child Benefit in full, we will move to a household rather than individual basis for assessing the tax charge.  

Neidle also said this was in the budget.

nikhilwoodruff:
Fixes #868


MaxGhenis (commented):
can we do this in the reform style instead? and add a reference for the contrib parameter?

nikhilwoodruff:
Added the structure change- but I'll hold off on a reference until the manifesto comes out and then use that.

PR #902, opened 2024-07-16
Add fiscal year targeting + SPI comparisons

### LINKED ISSUE #887: Move to fiscal years by default
The survey data years are all fiscal years, and I think in the short term we should move parameters that take effect on the 6th April back to 1st January- if we do this, this is all that's needed to say the model estimates fiscal year impacts.

This will enable direct comparisons with other models which we might want sooner.

In the long term we should figure out a better way but I think we should accelerate this process now.

nikhilwoodruff:
cc @anth-volk


policyengine_uk/data/datasets/spi.py
```diff
@@ -0,0 +1,78 @@
...
+            "household_id",
+        ]:
+            data[id_column] = data["person_id"]
+
+        data["state_id"] = np.ones(len(df), dtype=int)
+        data["person_state_id"] = data["state_id"]
+
+        data["household_weight"] = df.FACT
+        data["dividend_income"] = df.DIVIDENDS
+        data["gift_aid"] = df.GIFTAID
```
anth-volk, line 29: GIFTAID appears to be assigned both here and on line 65

nikhilwoodruff:
Tests passing- merging!

PR #900, opened 2024-07-15
Properly calculate dividend interest, Dividend Allowance, and Personal Allowance

anth-volk:
Fixes #896.

For a more in-depth explanation of the cases these changes cover, see the description in #896. This PR ensures that Personal Allowance and Dividend Allowance are properly calculated in cases where a portion of an individual's dividend income is covered by the Personal Allowance.

Note: This PR builds on top of #891, and the merging this PR will merge all of that PR's code changes, as well. This should be reviewed after that is merged to prevent duplication of review efforts.

PR #891, opened 2024-07-11
Make improvements to income tax-related variables

anth-volk:
Fixes #890 

This PR restructures many of the variables previously within the `income_tax` variable folder. It also restructures their corresponding tests. It adds reference material for many variables and changes the output type of `pays_scottish_income_tax` to Boolean.

This does not yet address potential issues with said variables, which will come in a later PR.

PR #903, opened 2024-07-16
Correctly calculate Personal Savings Allowance

### LINKED ISSUE #901: Savings Income does not factor in situations where Personal Allowance exempts a portion of income
Similar to #896. Personal Allowance and Savings Allowance currently don't function correctly if Personal Allowance covers some or all of an individual's savings income

anth-volk:
Fixes #901.

This is a very early-stage draft PR. More details to be added. This is based on #891 and #900, and requires the successful closure of both first before merging.


policyengine_uk/variables/gov/hmrc/income_tax/allowances/savings_allowance/savings_allowance.py
```diff
@@ -0,0 +1,47 @@
...
+    reference = "Income Tax Act 2007 s. 12B"
+    unit = GBP
+
+    def formula(person, period, parameters):
+        amounts = parameters(
+            period
+        ).gov.hmrc.income_tax.allowances.personal_savings_allowance
+
+        savings_allowance_category = person("savings_allowance_category", period)
+        max_allowance = amounts[savings_allowance_category.decode_to_str().lower()]
```
PavelMakarchuk, line 17: if we want to incorporate this type of structure, we need a breakdown parameter - we should consolidate all three parameters into one where each possible `savings_allowance_category` is defined with teh respective value

PavelMakarchuk: we then dont need the `.decode_to_str`

anth-volk:
Closing in light of #907

PR #931, opened 2024-08-22
Add mask to `private_school_vat`

### LINKED ISSUE #928: UK Household calculator crashing
UK app crashes with following error: "TypeError: Cannot read properties of null (reading 'concat')" after you enter household details and click on any of the impact tabs.

Occurs with all of the household detail combinations I tested so far.

Does not happen with the US app.

https://www.loom.com/share/0176a9dac9b04c0fa1bfddcc01c318f2?sid=5c525f85-0827-4dc2-9900-c790ca732c48

anth-volk:
Fixes #928 

PR adds a mask to `private_school_vat` to prevent crashing in cases where household weights aren't provided (e.g., household calculator runs).

This PR does not yet add tests for `private_school_vat` itself, which was previously untested. I will try to do so and add them to this PR, but considering the priority level of these fixes, wanted to give an option for quick merging.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    added:
+    - Mask applied to private_school_vat to prevent calculation error when household weights aren't provided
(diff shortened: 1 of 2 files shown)


MaxGhenis (approved):
Thanks - more reason for us to trim zero-weighted records out of the microdata, either in core or microdata production.

PR #935, opened 2024-08-27
Change Jupyter Book documentation build script

anmolchhabra21:
I should not use nbformat.read() because this does not recompile everything, so we’d lose out on model updates when running
 https://github.com/PolicyEngine/policyengine-app/issues/1429


anmolchhabra21:
image
Sample compilation time, for some of the notebooks using nbformat.

nikhilwoodruff:
Could you outline why this is needed? Thanks

nikhilwoodruff:
Closing due to inactivity.

PR #942, opened 2024-09-09
Add changes for `policyengine-uk-data`

nikhilwoodruff:


nikhilwoodruff:
I added 100 to the error tolerance for 2025 tests- these changed because I updated the CPI projection used to update parameters for future years.

Issue #972, opened 2024-10-22
Uprating benefit levels for 2025/26


vahid-ahmadi:
I updated the benefits parameters for 2025 and logged the changes in the database code.

nikhilwoodruff:
Could you share where this is, I don't see a pull request?

Also please see the amended title for an example for this issue.

PR #974, opened 2024-10-22
Uprating benefit levels for 2025/26

vahid-ahmadi:
I updated the benefits parameters for 2025 and logged the changes in the database code.


vahid-ahmadi:
Fixes #972

PR #977, opened 2024-10-23
Implement Household Income Rounding

vahid-ahmadi:
Added a function to calculate household income by summing market income and benefits, subtracting tax, and rounding the result to the nearest integer using numpy.


vahid-ahmadi:
Fixes #978

MaxGhenis:
It seems like this pr mostly adds uprating rather than the rounded income. Is there an issue for that? 

Also you need to tag the issue in the initial pr description to link it. It's still open.

PR #980, opened 2024-10-23
Updating values for uprating benefits 2025/26

vahid-ahmadi:
I reviewed and compared the 2023 and 2024 parameters to ensure accuracy, and then proceeded to update the values for 2025.


policyengine_uk/parameters/gov/benefit_uprating_cpi.yaml
```diff
@@ -1,6 +1,6 @@
 description: Most recent September CPIH index value, updated for each uprating occurrence (2014=100)
```
MaxGhenis: right?
```suggestion
description: Most recent September CPIH index value, updated for each uprating occurrence (2005=100)
```

nikhilwoodruff (approved):
LGTM on passing the latest tests.

PR #1030, opened 2025-02-10
Remove `calibration` folder in parameters

### LINKED ISSUE #1028: Remove `calibration` folder in parameters
This folder is hidden in the web app, has many descendants without metadata, and we've already agreed that this should be moved to the data repo (and the data repo already has these targets in a CSV).

### LINKED ISSUE #1029: Separate GitHub actions by trigger and use path filters
We should move to the format in https://github.com/policyengine/policyengine-us-data, where we apply path filters to GitHub actions to avoid rerunning tests or requiring version updates on changes that don't affect the package code.

nikhilwoodruff:
Fixes #1028
Fixes #1029


MaxGhenis:
Is policyengine-us-data also not using the policyengine-us calibration parameter folder? We just updated that from the new CBO forecast.

nikhilwoodruff:
It is- see [here](https://github.com/PolicyEngine/policyengine-us-data/blob/f8d6a50f13653421e1cffb8e3a25ec3f9c2181d5/policyengine_us_data/utils/loss.py#L181). Although I do think we should move to CSVs.

PR #1092, opened 2025-04-02
Some uprating fails on `ubuntu` but not `macos`

### LINKED ISSUE #1091: Some uprating fails on `ubuntu` but not `macos`
Found this after some debugging- on ubuntu, some benefit parameters don't uprate (income support and pension credit), while this works on mac.

nikhilwoodruff:
Fixes #1091

### DIFF
--- a/.github/workflows/pr_code_changes.yaml
+++ b/.github/workflows/pr_code_changes.yaml
@@ -21,7 +21,10 @@ jobs:
           with:
               args: ". -l 79 --check"
   Test:
-      runs-on: ubuntu-latest
+      strategy:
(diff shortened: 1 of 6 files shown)


nikhilwoodruff:
So after some debugging (thank you codespaces for making it easy to run a ubuntu instance), the issue occurs when uprating with a double-link (one parameter is uprated by another which is uprated by another).

nikhilwoodruff:
I've added a test that checks if both ubuntu and macos reach the exact same household net income number- I'll remove it after it passes because that target will change.

PR #1098, opened 2025-04-03
Correct ```is_parent``` variable

### LINKED ISSUE #1099: `is_parent` erroneously assigns parent status to children when parents are the same age
We should use `adult_index` instead.

nikhilwoodruff:
Fixes #1099


nikhilwoodruff:
Superseded by #1106

PR #1102, opened 2025-04-04
Test budgetary impacts for childcare programs

### LINKED ISSUE #1096: Add tests of budgetary impacts for childcare subsidies against government actuals
Some of these- extended childcare, for example, are well off (6 billion PE/1 billion OBR).

vahid-ahmadi:
Fixes #1096

How to use py code's draft: https://docs.google.com/document/d/1sNVE74VNQM1MnOa4erb-advlIkT0BS7jPsgk4Xr4uQE/edit?usp=sharing


nikhilwoodruff (changes requested):
This is different from how we usually test aggregates, and we don't have variables for totals elsewhere in the repo. Can you change this:
* Remove the total variable (it also erroneously calculates the total for each household rather than the total overall, so wouldn't work)
* Add a python pytest function that creates a microsimulation, sums universal entitlements, and asserts in bounds.

vahid-ahmadi:
We’re just getting the uk pr test passing and will move the take-up variable to the data package.

MaxGhenis:
Why not do this in calibration? We're going to remove microsimulation tests from this package at some point, and this deviates from the limited standard we currently apply.

nikhilwoodruff:
> Why not do this in calibration? We're going to remove microsimulation tests from this package at some point, and this deviates from the limited standard we currently apply.

I suggested we do this, because I think takeup is not enough to explain the 6x gap between e.g. extended childcare model and actuals. I think it's most efficient to get us close to the actual target with this as a draft PR (that probably involves changes to the UK model, or identifying bugs), and then add any changes to UK data that we need to.

vahid-ahmadi:
Order of priority:
1. If you are eligible for extended (you are working): you will go for extended (9 months to 4 years)
2. If you are not eligible for extended: you will go for universal (3-4 years old) and targeted (2 years old)

policyengine_uk/tests/microsimulation/test_childcare.py
```diff
@@ -69,6 +108,45 @@ def test_extended_childcare_entitlement_aggregate():
     )
 
 
+def test_extended_childcare_entitlement_recipients():
```
nikhilwoodruff, line 111: Something like this should definitely be wrapped in a function we call, we want to avoid code duplication

vahid-ahmadi:
Comparing DfE and PE spending and caseloads (link for [data](https://skillsfunding.service.gov.uk/view-latest-funding/national-funding-allocations/DSG/2024-to-2025), and [excel calculation](https://docs.google.com/spreadsheets/d/1HLwxCJAJQNHa64peQFfV47MuNqoOHBJ9lnFENXMTguE/edit?gid=1734460017#gid=1734460017)): 

image

PR #1105, opened 2025-04-04
Edit childcare programs' interactions

### LINKED ISSUE #1089: Check the childcare interactions
Working families can get either Tax-Free Childcare (up to £2000 per child, age 0–11) or Universal Credit Childcare (up to 85% of costs, age 0–16), alongside free childcare hours.
Universal, targeted, and extended childcare entitlements cannot be combined—families can only use one type at a time.

https://www.childcarechoices.gov.uk/combining-schemes

### LINKED ISSUE #1101: Set extended and targeted childcare entitlements to not be in addition to the universal entitlement
If we can't find a legislative reference for it (I couldn't after 15 mins) we still need to amend the parameters. This makes it clear ([Commons Library](https://commonslibrary.parliament.uk/free-childcare-entitlements-for-three-and-four-year-olds-in-england/)).

vahid-ahmadi:
Fixes #1089
Fixes #1101


vahid-ahmadi:
To avoid infinite loops, we need to establish a priority order for childcare programs. I assume the priority order is: universal > extended > targeted.

@nikhilwoodruff What do you think?

nikhilwoodruff:
Can you add the source for these incompatibilities? I'm concerned by the requirement that you can't claim universal childcare with targeted. Saying that universal childcare disqualifies you might kill off the whole program, no?

PR #1139, opened 2025-05-31
One variable per file

### LINKED ISSUE #1027: One variable per file
This should be enforced everywhere.

MaxGhenis:
Fix #1027


policyengine_uk/tests/microsimulation/test_validity.py
```diff
@@ -1,13 +1,13 @@
-from policyengine import Simulation
+from policyengine_uk import Microsimulation
```
MaxGhenis: @nikhilwoodruff do we intend to use policyengine.py in policyengine-uk at this point?

MaxGhenis: never mind left it in

nikhilwoodruff:
This is quite a big PR so I'm going to run the EFRS before and after and assert no difference in any household.

nikhilwoodruff:
Verified!

image

nikhilwoodruff (approved):

PR #1205, opened 2025-07-11
Switch from per capita employment income to average earnings forecast

### LINKED ISSUE #1204: Use OBR's average earnings growth forecast instead of per capita employment income

janton10:
Fixes #1204


nikhilwoodruff (approved):
I am OK with this, though we should check with the microsim how it affects an example reform or a projection of real incomes (I think this probably causes a slight drop in accuracy for e.g. income tax revenue raisers until we add future year calibration).

nikhilwoodruff:
@janton10 all PRs to policyengine-uk need a changelog update file (I'll add a PR test for this) to update the package. Example [here](https://github.com/PolicyEngine/policyengine-uk/pull/1200/files#diff-5885042762833a4aef51405a4036067e4e6aefa7a5091a2987d1d9d1fc2647e8), I'll add for this one.

janton10:
Not convinced it'll cause a drop in accuracy (albeit you'd have to recalibrate the weights of course). Calculations are very similar but average earnings growth forecast uses employment forecast (see OBR footnotes Table 1.6)

MaxGhenis:
@nikhilwoodruff why do you expect an accuracy drop from the switch?

nikhilwoodruff:
> @nikhilwoodruff why do you expect an accuracy drop from the switch?

Mostly a guess I suppose, but we're no longer effectively guaranteed to hit the projections for aggregate employment income.

nikhilwoodruff:
I'm going to add a basket of reform tests to this so we can use it moving forward.

nikhilwoodruff (approved):

nikhilwoodruff:
... ubuntu returning different values to macos, investigating.

nikhilwoodruff (changes requested):
Not to merge til I fix the tests

nikhilwoodruff:
Filed https://github.com/PolicyEngine/policyengine-uk/issues/1209 and temporarily removing the ubuntu action so as to unblock this. Will keep working on establishing why ubuntu is different after merging.

PR #1208, opened 2025-07-11
Fix basic state pension calculation

### LINKED ISSUE #1090: Check basic state pension logic
It seems ```amount_in_data_year``` doesn’t depend on ```maximum_basic_sp```. It never goes to zero when assigning ```maximum_basic_sp``` to zero.

https://gist.github.com/vahid-ahmadi/dda76abf28799016491af40b474d0828

Image

vahid-ahmadi:
Fixes #1090


vahid-ahmadi:
The original issue was that the ```basic_state_pension``` formula returned ```0``` when ```simulation.dataset``` was None, preventing the calculation from ever reaching the ```amount_in_data_year``` logic where ```maximum_basic_sp``` would be applied. I fixed this by modifying the formula to handle the no-dataset case by using the current period for calculations instead of returning 0 early, ensuring that ```amount_in_data_year``` is always calculated and properly depends on ```maximum_basic_sp``` through the ```min_(reported, maximum_basic_sp)``` operation.

policyengine_uk/tests/policy/baseline/gov/dwp/basic_state_pension.yaml
```diff
@@ -0,0 +1,44 @@
...
+- name: Non-basic state pension not affected by basic pension maximum
+  period: 2021
+  input:
+    people:
+      person1:
+        state_pension_reported: 5000
+        state_pension_type: NEW  # Not basic pension
+        age: 70
+  output:
+    basic_state_pension: 5000  # Should not be affected by basic pension maximum
```
nikhilwoodruff: Why is this 5000 when the state pension type is NEW?

nikhilwoodruff:
@policyengine can you review this/fix issues and get enough done to get this either over the line, or closed if not worth continuing. Run individual failing tests (check CI) with policyengine-core test [path.yaml] -c policyengine_uk to avoid big memory usage and wait times

nikhilwoodruff:
Can you sort the merge conflicts @policyengine

nikhilwoodruff (approved):

policyengine_uk/tests/policy/baseline/gov/dwp/basic_state_pension.yaml
nikhilwoodruff: fine as a unit test value

PR #1223, opened 2025-07-14
Fix parameter uprating bug

### LINKED ISSUE #1209: Ubuntu is returning different results from macos
Identified by https://github.com/PolicyEngine/policyengine-uk/pull/1205.

I've made a private GitHub repo [here](https://github.com/nikhilwoodruff/policyengine-uk-private) which runs code on both ubuntu-latest and macos-latest that stores the full entity dataset as artifacts so I can see exactly which variables are different (have so far established the input data is the exact same, not modified by OS behaviour).

vahid-ahmadi:
Fixes #1209


nikhilwoodruff (changes requested):
Thanks for taking a look. Could you add a test for this? We can reinstate the ubuntu-latest runner, if you add it it should pass/fail on the same tests.

vahid-ahmadi:
> Thanks for taking a look. Could you add a test for this? We can reinstate the ubuntu-latest runner, if you add it it should pass/fail on the same tests.

Thanks! Added.

nikhilwoodruff:
unless you want to keep this @vahid-ahmadi , ok if I close it? not actually sure that the proxy object works, and I think it adds a lot of bloat

PR #1238, opened 2025-07-15
Rename all variables and functions to lowercase

### LINKED ISSUE #259: Lowercase function and variable names
We have a number of variable names that are partly uppercase, referring to acronymized benefit programs like PIP. Yet [PEP8 makes no concessions for acronyms](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names), saying:

>Function names should be lowercase, with words separated by underscores as necessary to improve readability.
>Variable names follow the same convention as function names.

This would be a substantive change but I think we should consider the precedent we're setting across PolicyEngine software, including openfisca-us.

juaristi22:
Fix #259


policyengine_uk/entities.py
```diff
@@ -19,7 +19,7 @@
     containing_entities=["state"],
 )
 
-BenUnit = build_entity(
+ben_unit = build_entity(
```
MaxGhenis: undo

policyengine_uk/variables/gov/dwp/CTC_child_element.py
```diff
@@ -12,14 +12,14 @@ class CTC_child_element(Variable):
     def formula(benunit, period, parameters):
         person = benunit.members
         CTC = parameters(period).gov.dwp.tax_credits.child_tax_credit
```
vahid-ahmadi: ctc

policyengine_uk/variables/gov/dwp/CTC_severely_disabled_child_element.py
```diff
@@ -12,12 +12,12 @@ class CTC_severely_disabled_child_element(Variable):
...
+        is_child_for_ctc = person("is_child_for_ctc", period)
         is_severely_disabled_for_benefits = person(
             "is_severely_disabled_for_benefits", period
         )
         is_severely_disabled_child = (
-            is_child_for_CTC & is_severely_disabled_for_benefits
+            is_child_for_ctc & is_severely_disabled_for_benefits
         )
         severely_disabled_children = benunit.sum(is_severely_disabled_child)
         CTC = parameters(period).gov.dwp.tax_credits.child_tax_credit
```
vahid-ahmadi: ctc

policyengine_uk/variables/gov/dwp/ctc_child_limit_affected.py
```diff
@@ -10,16 +10,16 @@ class ctc_child_limit_affected(Variable):
     def formula(benunit, period, parameters):
         person = benunit.members
         CTC = parameters(period).gov.dwp.tax_credits.child_tax_credit
```
vahid-ahmadi: ctc

vahid-ahmadi (changes requested):

policyengine_uk/variables/gov/dwp/CTC_child_element.py
MaxGhenis: if already renaming this, change to `p` for consistency with US standards

policyengine_uk/variables/gov/dwp/ctc_child_limit_affected.py
juaristi22: aaa yess will do a more thorough sweep

nikhilwoodruff:
Thank you!! Though we need to be **super** careful with this as it has high potential to break things. Luckily most of the important variables are already lowercase, but lets just make sure by running a notebook microsim, testing with policyengine.py etc.

vahid-ahmadi (approved):

nikhilwoodruff:
@policyengine can you fix all these merge conflicts (also make the folders lowercase, might be tricky with git)?

nikhilwoodruff:
@policyengine look at the test output, we've got issues. also, could you add a small intercept in Simulation.calculate (and Microsimulation.calculate) to make sure we convert the variable name to lowercase, just to be backwards compatible?

nikhilwoodruff:
@policyengine continue

nikhilwoodruff:
@policyengine you missed some test yaml bits

nikhilwoodruff:
@policyengine tests fail- circular reference on rent?

nikhilwoodruff:
@policyengine still the same issue. instead of stopping when done, wait to see if CI passes

nikhilwoodruff:
@policyengine nearly there, just pass the tests finally

MaxGhenis:
Closing this PR for now - we may revisit the lowercase renaming in the future with a fresh approach. Thanks to María for the initial work on this.

PR #1246, opened 2025-07-17
Use outturn data for council tax growth, 2023-25

### LINKED ISSUE #1243: Use published CT values for Eng/Scot/Wales for current year

janton10:
#1243


nikhilwoodruff (approved):
Should merge as-is but in future you don't actually need to run make changelog, the GH actions do it. Just the changelog_entry.yaml file is enough (this also means that when we have concurrent PRs, we don't have to redo all the changelog/version numbers if we merge in a different order than expected)

janton10:
> Should merge as-is but in future you don't actually need to run make changelog, the GH actions do it. Just the changelog_entry.yaml file is enough (this also means that when we have concurrent PRs, we don't have to redo all the changelog/version numbers if we merge in a different order than expected)

Makes sense!

PR #1296, opened 2025-08-03
Add OBR labour supply response elasticities

### LINKED ISSUE #938: OBR elasticity breakdown
More detailed than CBO's: https://obr.uk/docs/dlm_uploads/NICS-Cut-Impact-on-Labour-Supply-Note.pdf

image

image

nikhilwoodruff:
Consolidating the labour supply response logic here and standardising to OBR assumptions.

Fixes #938


nikhilwoodruff:
There is a question here about how to handle the participation response, since the OBR does this on an aggregate level:

image

I think the method that is technically equivalent, but is compatible with our microdata-first approach would be to:

1. Assign participation elasticities from participation groups as in the OBR doc
2. Compute 'gain to work' for each person
3. For each group (participation elasticity unique groups), calculate the 'extra participation' effect by multiplying elasticity by relative change in 'gain to work' and summing (after this, we will have e.g. +2m [lone parents youngest child aged 0-2], etc)
4. For each group, find the out of work population, and randomly take e.g. 2m people from there, assign them average hours and earnings for the in-work population for that group.

Welcoming throughts @MaxGhenis @uta-bolt @janton10. @MaxGhenis I know we should probably just train a quantile regression forest model to predict the distribution of earnings+hours for 4, but that would be deviating from the OBR assumptions, and I think it's useful in itself to have a complete match on the assumptions (I think this is mathematically equivalent)

MaxGhenis:
Does OBR describe how they bring this back into a microsim (if at all)? We need to capture the lumpiness of the response.

Your proposal could move individuals into work even if their own return to work is unchanged, just because their group was affected. An alternative would be estimating the probability that each moves into work based on the elasticity, then randomly assigning based on that.

janton10:
> For each group, find the out of work population, and randomly take e.g. 2m people from there, assign them average hours and earnings for the in-work population for that group.

"The participation effect increases the number of people in employment by 28,000 2028-29, as a result of the increase in weekly ‘gain-to-work’ of £8.587 on average. However, **because people join the labour force at below average hours**, the increase in labour supply from the participation effect is only 14,000 in full-time equivalent terms." https://obr.uk/docs/dlm_uploads/NICS-Cut-Impact-on-Labour-Supply-Note.pdf#page=7

Don't think your method from above captures this.

nikhilwoodruff:
> Does OBR describe how they bring this back into a microsim (if at all)? We need to capture the lumpiness of the response.
> 
> Your proposal could move individuals into work even if their own return to work is unchanged, just because their group was affected. An alternative would be estimating the probability that each moves into work based on the elasticity, then randomly assigning based on that.

OBR doesn't, because I don't think they do bring this back into a microsim.

MaxGhenis:
How do they determine the hours for entrants?

And without microsim how do they estimate the budgetary impact?

nikhilwoodruff:
> > For each group, find the out of work population, and randomly take e.g. 2m people from there, assign them average hours and earnings for the in-work population for that group.
> 
> "The participation effect increases the number of people in employment by 28,000 2028-29, as a result of the increase in weekly ‘gain-to-work’ of £8.587 on average. However, **because people join the labour force at below average hours**, the increase in labour supply from the participation effect is only 14,000 in full-time equivalent terms." https://obr.uk/docs/dlm_uploads/NICS-Cut-Impact-on-Labour-Supply-Note.pdf#page=7
> 
> Don't think your method from above captures this.

Right yes. So I think I need to follow (@MaxGhenis on your question two):

> **Note: We assume that those entering the labour market work 18.8 hours per week, which was the average number of hours worked by part-time workers in 2021 according to ASHE data.**

And then it should match?

nikhilwoodruff:
> How do they determine the hours for entrants?

See above, 18.8 hours per week.

> And without microsim how do they estimate the budgetary impact?

I assume they estimate changes to revenues from increased employment incomes etc, then subtract this from the original costing, but will dig more.

nikhilwoodruff:
> Your proposal could move individuals into work even if their own return to work is unchanged, just because their group was affected. An alternative would be estimating the probability that each moves into work based on the elasticity, then randomly assigning based on that.

I think we would need to impute the target wage for this no? And would deviate from the OBR assumptions.

MaxGhenis:
>I assume they estimate changes to revenues from increased employment incomes etc, then subtract this from the original costing, but will dig more.

That might work for taxes but benefit changes from participation responses will vary significantly within the groups.

MaxGhenis:
>I think we would need to impute the target wage for this no? And would deviate from the OBR assumptions.

We definitely need the wage even for aggregate analysis if estimating revenue changes.

nikhilwoodruff:
> > I think we would need to impute the target wage for this no? And would deviate from the OBR assumptions.
> 
> We definitely need the wage even for aggregate analysis if estimating revenue changes.

Sorry, to clarify: the OBR's approach seems equivalent to assigning average wages and 18.8 hours per week for each of the groups. Following that here

nikhilwoodruff:
> > I assume they estimate changes to revenues from increased employment incomes etc, then subtract this from the original costing, but will dig more.
> 
> That might work for taxes but benefit changes from participation responses will vary significantly within the groups.

Agree but we need to be able to match OBR assumptions. We can provide alternatives

nikhilwoodruff:
Success criteria for merging (will add a test): reproduce the original labour supply response from the Autumn 2023 NICs cut, within 25% (different economy growthfactors since then).

nikhilwoodruff:
OK, we are currently lower at 48k vs 74k for the progression response. Digging into why. We match the average marginal net wage increase of 24p (OBR: 23p to nearest penny), and £13.70 (OBR: £13 to nearest pound).

nikhilwoodruff:
I wasn't following the right methodology: hours changed under the OBR definition should be adjusted for each employee's wage, rather than the average wage. Under that, we get 68,000 FTE equivalents compared to the OBR's 79,000. That seems close enough, so I'm going to move onto the participation elasticity response.

nikhilwoodruff:
@MaxGhenis as part of this I'm doing most of this in a custom set of functions because they make debugging way easier: returning a dataframe of per-individual changes. Will think about how to best integrate this into formulas. 
image

nikhilwoodruff:
*Aggregate* participation model now in, and is close to the OBR's estimate. Recapping current outputs below:

| Metric | OBR | PolicyEngine | Difference | Relative difference |
|--------|-----|--------------|------------|---------------------|
| **Participation impact** |
| Employment (headcount) | 28,000 | 25,122 | -2,878 | -10.3% |
| Full-time equivalent | 14,000 | 12,595 | -1,405 | -10.0% |
| **Progression impact** |
| Full-time equivalent | 79,000 | 67,000 | -12,000 | -15.2% |
| **Total effect** |
| Full-time equivalent | 94,000 | 79,595 | -14,405 | -15.3% |

nikhilwoodruff:
Added also a standard output tool to produce the below for any reform. Latest outputs:

| Metric | OBR | PolicyEngine | Difference | Relative difference |
|--------|-----|--------------|------------|---------------------|
| **Participation impact** |
| Employment (headcount) | 28,000 | 30,391 | +2,391 | +8.5% |
| Full-time equivalent | 14,000 | 15,236 | +1,236 | +8.8% |
| **Progression impact** |
| Full-time equivalent | 79,000 | 69,607 | -9,393 | -11.9% |
| - Substitution effect | - | 77,389 | - | - |
| - Income effect | - | -7,791 | - | - |
| **Total effect** |
| Full-time equivalent | 94,000 | 84,843 | -9,157 | -9.7% |
| Total hours change | 0.30% | 0.27% | -0.03pp | -10.0% |

nikhilwoodruff:
Will revisit this PR after the uk data update and see how it changes.

nikhilwoodruff:
Updated outputs:

Metric | OBR | PolicyEngine | Difference | Relative difference
-- | -- | -- | -- | --
Participation impact |   |   |   |  
Employment (headcount) | 28,000 | 8,296 | -19,704 | -70.4%
Full-time equivalent | 14,000 | 4,159 | -9,841 | -70.3%
Progression impact |   |   |   |  
Full-time equivalent | 79,000 | 77,587 | -1,413 | -1.8%
Total effect |   |   |   |  
Full-time equivalent | 94,000 | 81,745 | -12,255 | -13.0%

Right, so the participation effect is not as strong when only applying the elasticities of 'moving into work' onto nonworkers and vice versa (in this approach, we are effectively throwing away any strengthening of work incentives among workers and vice versa). I'm going to try adjusting (as a last resort, will say this is done and move on if it doesn't work) the participation logic to take the overall relative gain to work change (mean relative change in gain to work, for each group) and ensure that this overall mean change is matched when zeroing out the gain to work rel changes in the wrong directions.

nikhilwoodruff:
Updated results: overall closer, but mostly because of progression not participation. Going to mark this as an OK replication for now.

Metric | OBR | PolicyEngine | Difference | Relative difference
-- | -- | -- | -- | --
Participation impact |   |   |   |  
Employment (headcount) | 28,000 | 13,862 | -14,138 | -50.5%
Full-time equivalent | 14,000 | 6,950 | -7,050 | -50.4%
Progression impact |   |   |   |  
Full-time equivalent | 79,000 | 77,106 | -1,894 | -2.4%
Total effect |   |   |   |  
Full-time equivalent | 94,000 | 84,056 | -9,944 | -10.6%

PR #1327, opened 2025-08-28
Fix behavioral response calculations returning zero FTE impacts

### LINKED ISSUE #1328: Debug behavioral response calculations returning zero FTE impacts
The behavioral response implementation in ```labour_supply.py``` was systematically returning zero FTE impacts despite valid policy reforms with significant income effects. This affected the accuracy of dynamic scoring for tax and benefit policy changes.

vahid-ahmadi:
Fixes #1328

**Problem**

The behavioral response implementation in ```labour_supply.py``` was systematically returning zero FTE impacts despite valid policy reforms with significant income effects. This affected the accuracy of dynamic scoring for tax and benefit policy changes.

**Root Causes**

1. Simulation state corruption: ```sim.reset_calculations()``` was clearing all computed values, causing subsequent calculations like ```household_net_income``` to return NaN or £0
2. NaN propagation: Division by zero in wage calculations created NaN values that weren't properly handled, breaking elasticity calculations
3. Timing issues: Income changes were calculated after simulation state corruption occurred

**Solution**

- Removed ```sim.reset_calculations()```: Prevents simulation state corruption while still applying behavioral responses correctly
- Enhanced NaN handling: Added ```.fillna(0)``` and included np.nan in replacement list for wage calculations
- Pre-calculated income changes: Calculate income effects before any simulation modifications to avoid timing issues

**Impact**

- Behavioral responses now work correctly, showing realistic FTE impacts (e.g., 51,197 total FTEs for major reforms)
- All existing tests pass (569/569) - no regressions introduced
- Dynamic scoring functionality restored for accurate policy impact assessment

**Files Changed**

- ```policyengine_uk/dynamics/labour_supply.py``` - Core behavioral response fixes


policyengine_uk/dynamics/labour_supply.py
```diff
@@ -126,6 +126,24 @@ def apply_labour_supply_responses(
     if (not follow_obr) or (sim.baseline is None):
         return
 
+    # Calculate income changes using household_net_income
+    baseline_income = sim.baseline.calculate(
+        target_variable, year, map_to="person"
+    )
+    reform_income = sim.calculate(target_variable, year, map_to="person")
+
+    if hasattr(baseline_income, "values"):
```
nikhilwoodruff: This shouldn't be needed

nikhilwoodruff:
This isn't the fix- because the web app doesn't use this code, it uses the old behavioural response elasticity variables (we just use this function for custom Python analysis)

vahid-ahmadi:
> This isn't the fix- because the web app doesn't use this code, it uses the old behavioural response elasticity variables (we just use this function for custom Python analysis)

Thanks for the clarification. Yes, this fix doesn't affect the web app's behavioral response. I'm taking a look at that as well.

However, these changes fix the labour supply response functionality for the Python code. With this fix, you can get the results running the [examples](https://policyengine.github.io/policyengine-uk/dynamics/).

PR #1330, opened 2025-08-29
Fix/incidence bug

janton10:


nikhilwoodruff (approved):
Weird bug, but thanks!

PR #1351, opened 2025-10-02
Fix age progression in multi-year projections for two-child limit

MaxGhenis:
## Summary

Fixes the bug where the two-child limit budgetary impact was incorrectly constant across years (e.g., same £2.2bn cost in 2026 and 2029).

## Problem

The uprating system in `economic_assumptions.py` was updating monetary values (income, consumption, etc.) but **not aging the population**. This caused:

- Ages to stay frozen at base year values
- Birth years to drift incorrectly: `birth_year = period.year - age` with static age means birth year changes each year
- Number of children subject to two-child limit to remain constant over time

Since children born before April 2017 are exempt from the two-child limit, the cost of repealing it should increase over time as:
- Pre-2017 children age out of the system
- More post-2017 children enter the population

## Solution

Added age progression to `apply_single_year_uprating()` which increments ages by 1 for each projection year.

Since `birth_year` is calculated as `period.year - age`, incrementing ages keeps birth years constant:
- 2024: age=7, birth_year = 2024-7 = 2017 ✓
- 2026: age=9, birth_year = 2026-9 = 2017 ✓ (now correct!)
- 2029: age=12, birth_year = 2029-12 = 2017 ✓ (now correct!)

## Test Plan

- [x] Created unit tests in `test_age_progression.py` documenting expected behavior
- [x] Tests pass locally
- [x] Changelog updated

## Impact

This fix ensures that:
1. The two-child limit cost projections are now accurate and increase over time
2. All policies with age-based criteria work correctly in multi-year projections
3. Birth year calculations remain stable across projection years

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), 
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.55.0] - 2025-10-02 14:18:49
+
+### Fixed
+
+- Fix age progression in multi-year projections to properly age the population over time, resolving issue where two-child limit budgetary impact was incorrectly constant across years
+
 ## [2.54.0] - 2025-09-30 14:40:31
 
 ### Added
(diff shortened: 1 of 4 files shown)


MaxGhenis:
closing, this is incorrect

PR #1355, opened 2025-10-05
Make country package purely deterministic - read stochastic variables from dataset

MaxGhenis:
## Summary

This PR removes all random number generation from policyengine-uk. All stochastic variables are now generated in policyengine-uk-data and read from the dataset. The country package is now a purely deterministic rules engine.

**⚠️ MERGE ORDER: The companion policyengine-uk-data PR #203 must be merged FIRST, then this PR**

### Removed
- All take-up rate parameters (moved to policyengine-uk-data)
- All random seed/draw variables for take-up decisions
- Random seed variables for other stochastic processes

### Simplified
All would_claim variables now use dataset values with deterministic fallbacks:
- would_claim_child_benefit (default: True)
- child_benefit_opts_out (default: False)
- would_claim_pc (default: True)
- would_claim_uc (default: True)
- would_claim_tfc (default: True)
- would_claim_extended_childcare (default: True)
- would_claim_universal_childcare (default: True)
- would_claim_targeted_childcare (default: True)

Other stochastic variables simplified to dataset-only:
- household_owns_tv (default: True)
- would_evade_tv_licence_fee (default: False)
- main_residential_property_purchased_is_first_home (default: False)

### Added
- would_claim_marriage_allowance variable (default: True)
- higher_earner_tie_break variable (default: 0.0) - for tie-breaking
- attends_private_school_random_draw variable (default: 1.0) - for income-conditional

### Preserved formulas (fully deterministic)
These variables keep their formulas but with NO random() calls:
- attends_private_school (uses random draw from dataset)
- is_disabled_for_benefits (deterministic rule: disabled if receives qualifying benefits)
- is_higher_earner (uses random draw from dataset for tie-breaking)

### Test updates
Updated expected fiscal impacts in `reforms_config.yaml` to reflect the new stochastic simulation method.

## Trade-offs

**IMPORTANT**: Take-up rates can no longer be adjusted dynamically via policy reforms or in the web app. They are fixed in the microdata. This is an acceptable trade-off for the cleaner architecture of keeping the country package purely deterministic.

To adjust take-up rates for analysis, the microdata must be regenerated with updated parameter values in policyengine-uk-data.

## Test Plan

- [x] Package imports successfully
- [x] All existing tests pass
- [x] Microsimulations produce correct results
- [x] Policy calculator (non-microsim) still works with deterministic defaults

## Related PRs

- policyengine-uk-data: #203 (must be merged FIRST)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #1359, opened 2025-10-15
Fix randomness between runs caused by UC reform

### LINKED ISSUE #1358: Randomness between runs caused by UC reform
Running the same calculation multiple times was producing different results due to non-deterministic behaviour in the UC reform logic. This needs fixing to ensure reproducible outputs.

nikhilwoodruff:
The UC reform logic was using numpy's global random state, which caused different results on each run. This changes it to use a seeded random number generator for reproducible outputs.

Closes #1358

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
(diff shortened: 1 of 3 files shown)


MaxGhenis:
See also #1355 for a more general solution

PR #1427, opened 2025-12-01
Add owns vehicle variable

vahid-ahmadi:
Summary

Adds a new owns_vehicle input variable to support calibration of vehicle ownership rates in policyengine-uk-data.

  This addresses https://github.com/PolicyEngine/policyengine-uk-data/issues/236, which identified that the  microdata shows 54.1% of households with zero fuel spending, while NTS 2024 data shows only 22% of UK households lack a car - a 32 percentage point gap.

  Changes

  - Added owns_vehicle boolean input variable at household level

  How it works

This is an input variable (no formula). The value will be imputed in policyengine-uk-data based on LCFS fuel spending data (households with petrol/diesel spending are assumed to own a vehicle), then calibrated to NTS 2024 targets (~78% vehicle ownership).

  This approach:
  - Keeps the relationship between vehicle ownership and fuel spending in the data layer, not the tax-benefit system
  - Allows proper calibration/reweighting
  - Handles EVs correctly (they can be assigned owns_vehicle=True without fuel spending)

  References

  - https://www.gov.uk/government/statistics/national-travel-survey-2024/nts-2024-household-car-availability-and-trends-in-car-trips
  - https://assets.publishing.service.gov.uk/media/6781339100e3d719f19217f1/dft-car-ownership-evidence-review.pdf


vahid-ahmadi:
We need to change some tests from other programs. Why?:

The issue is that adding a new variable changes the order in which variables are loaded, which affects the random seed used by random(). This is a known quirk of policyengine_core - the random() function uses entity IDs as seeds, and adding new variables can shift the random sequence.

MaxGhenis (changes requested):
should be in assets not consumption, and number of vehicles

vahid-ahmadi:
> should be in assets not consumption, and number of vehicles

Moved them to the wealth folder (there's no assets folder).

PR #1434, opened 2025-12-03
[Draft] Fix fiscal year parameter handling (on-the-fly approach - needs work)

MaxGhenis:
## Summary

- Fix fiscal year parameter handling to use April 30 reference date for annual queries
- Ensures policies that change on April 6 (UK fiscal year start) are correctly reflected in simulations

## Problem

Previously, `convert_to_fiscal_year_parameters()` only covered years 2015-2025 (`range(2015, 2026)`). When querying parameters for year 2026 or later, the system fell back to January 1 as the reference date instead of April 30.

This caused policies that change on April 6 (UK fiscal year start) to not be reflected in annual simulations. For example:
- The two-child limit repeal on April 6, 2026 would show as `limit=2` for 2026 simulations instead of `limit=infinity`
- The salary sacrifice cap starting April 6, 2029 would not be reflected in 2029 simulations

## Solution

1. Add `convert_instant_to_fiscal_year()` function that converts year-only and January 1 queries to April 30 of that year
2. Override `get_parameters_at_instant()` in `CountryTaxBenefitSystem` to apply this conversion on-the-fly
3. Make `convert_to_fiscal_year_parameters()` a no-op (deprecated) since the conversion is now done on-the-fly

This ensures:
- `param("2026")` returns the April 30, 2026 value (fiscal year 2026/27)
- Specific date queries like `param("2026-04-05")` are unchanged  
- Works for any future year without needing to extend a range

## Test plan

- [x] Run new fiscal year parameter tests
- [x] Verify existing tests pass
- [ ] CI passes

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,5 +1,4 @@
 - bump: minor
   changes:
-    added:
-      - Two child limit repeal from April 2026 (Autumn Budget 2025) - sets UC and Tax Credits child element limit to infinity
-      - Salary sacrifice pension cap of £2,000 from April 2029 (Autumn Budget 2025)
+    fixed:
+      - Fix fiscal year parameter handling to use April 30 reference date for annual queries, ensuring policies that change on April 6 (UK fiscal year start) are correctly reflected in simulations.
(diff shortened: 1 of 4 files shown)


MaxGhenis:
Closing this stale draft — the fiscal year conversion approach needs a fresh design. See #1436 for the tracking issue.

PR #1440, opened 2025-12-03
Bump policyengine-core to 3.23.0

MaxGhenis:
## Summary

Updates policyengine-core from >=3.19.3 to >=3.23.0.

## Why

policyengine-core 3.23.0 adds strict enum validation - invalid enum values now raise `ValueError` instead of silently converting to index 0 (which caused silent data corruption).

This PR tests that all enum values in our test files are valid. If CI fails, it means we have invalid enum values that need to be fixed.

## Related

- policyengine-core#411 - The fix for silent enum data corruption

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+      - Bump policyengine-core to 3.23.0 (adds strict enum validation).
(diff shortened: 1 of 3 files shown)


MaxGhenis:
The lint failure has been fixed. The test failure is a **pre-existing issue** unrelated to the core bump - the VAT reform impact test expected 18.7 billion but now calculates 22.1 billion. This may be due to updated microdata or parameter changes, but it's not caused by the enum validation changes in policyengine-core 3.23.0.

The enum validation is working correctly - no invalid enum values were detected in the UK test suite.

PR #1494, opened 2026-01-28
Relax Python version requirement to >=3.10

### LINKED ISSUE #1493: Relax Python version requirement from >=3.13 to >=3.10
## Problem

`pyproject.toml` currently requires `>=3.13,<3.14`, but this appears to be unintentional:

1. The old `setup.py` had `python_requires=">=3.7"`
2. `policyengine-us` uses `>=3.10,<3.14`
3. `policyengine-core` uses `>=3.10`
4. The classifiers in pyproject.toml list Python 3.10, 3.11, 3.12, 3.13

When pyproject.toml was introduced in PR #1395 (Nov 2025), it was set to `>=3.13` rather than matching the previous `>=3.7` or aligning with other PE packages.

## Impact

Users on Python 3.10-3.12 get a confusing pip error:
```
ERROR: Could not find a version that satisfies the requirement policyengine-uk==2.72.3
```

This particularly affects Modal deployments and other environments that default to Python 3.11.

## Proposed Fix

Change in `pyproject.toml`:
```toml
requires-python = ">=3.10,<3.14"
```

This aligns with policyengine-us and should work unless there's a specific 3.13+ dependency I'm not aware of.

MaxGhenis:
## Summary
- Changes `requires-python` from `>=3.13,<3.14` to `>=3.10,<3.14`

Fixes #1493

## Context
The `>=3.13` requirement was unintentionally set when migrating from `setup.py` (which had `>=3.7`) to `pyproject.toml`. This aligns with policyengine-us which uses `>=3.10,<3.14`.

## Test plan
- [ ] CI passes on Python 3.10, 3.11, 3.12, 3.13

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
     "Topic :: Scientific/Engineering :: Information Analysis",
 ]
-requires-python = ">=3.13,<3.14"
+requires-python = ">=3.10,<3.14"
 dependencies = [
     "policyengine-core>=3.23.6",
     "microdf-python>=1.2.1",


MaxGhenis:
Closing as obsolete. The repo has since moved in the opposite direction: current `pyproject.toml` requires Python 3.13, so relaxing to `>=3.10` is no longer aligned with mainline support.

PR #1507, opened 2026-02-23
Fix UC standard allowance reforms being ignored (#1472)

### LINKED ISSUE #1472: Reforms to the Universal Credit standard allowances do not work
This is because the rebalancing uplift directly modifies and saves the standard allowance variable values, bypassing their formula.

MaxGhenis:
## Summary
- UC standard allowance reforms now work correctly
- Moved rebalancing uplift logic from `set_input()` in `uc_reform.py` into the `uc_standard_allowance` formula itself
- User reforms to `gov.dwp.universal_credit.standard_allowance.amount` are now respected

## Root cause
The UC rebalancing scenario used `sim.set_input("uc_standard_allowance", year, ...)` during `Simulation.__init__()`, which baked values BEFORE user reforms were applied. The formula was never called, so parameter reforms were silently ignored.

## Fix
The formula now reads the rebalancing parameters (`rebalancing.active` and `rebalancing.standard_allowance_uplift`) and applies the uplift multiplicatively to the base amount. Since the formula always runs, user reforms to the base amount are respected.

## Test plan
- [x] New test: reform doubling UC SA amount produces doubled output
- [x] All existing tests pass

Fixes #1472

### DIFF
--- a/changelog.d/fix-uc-standard-allowance-reform.fixed.md
+++ b/changelog.d/fix-uc-standard-allowance-reform.fixed.md
@@ -1,0 +1,1 @@
+Fix UC standard allowance reforms being ignored due to rebalancing set_input.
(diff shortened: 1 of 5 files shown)


MaxGhenis:
Superseded by #1547, which reapplied this fix on current main, passed fresh CI, and was merged. Closing the stale conflicting branch.

PR #1508, opened 2026-02-23
Fix abolishing council tax having no budget impact (#1153)

### LINKED ISSUE #1153: Abolish Council Tax has no budgetary impact

MaxGhenis:
## Summary
- Abolishing council tax now correctly shows budget impact
- Created `council_tax_applicable` variable that returns 0 when `abolish_council_tax` is True
- Replaced `council_tax` with `council_tax_applicable` in all downstream aggregation variables
- Removed special-case `abolish_council_tax` formula overrides (now unnecessary)

## Root cause
`council_tax` was a pure input variable with no formula. The `abolish_council_tax` parameter removed it from aggregation sums in both `household_tax` AND `pre_budget_change_household_tax`, so the reform-vs-baseline delta was always zero.

## Fix
New `council_tax_applicable` variable: returns 0 if abolished, otherwise returns the `council_tax` input. All downstream variables (`gov_tax`, `household_tax`, `pre_budget_change_household_tax`, `hbai_household_net_income`, `household_benefits`) now reference `council_tax_applicable` instead.

## Test plan
- [x] New test: abolishing council tax increases household_net_income
- [x] All existing tests pass

Fixes #1153

### DIFF
--- a/changelog.d/fix-abolish-council-tax.fixed.md
+++ b/changelog.d/fix-abolish-council-tax.fixed.md
@@ -1,0 +1,1 @@
+Fix abolishing council tax having no budget impact.
--- a/policyengine_uk/tests/microsimulation/reforms_config.yaml
+++ b/policyengine_uk/tests/microsimulation/reforms_config.yaml
@@ -16,7 +16,7 @@ reforms:
   parameters:
     gov.hmrc.child_benefit.amount.additional: 25
 - name: Reduce Universal Credit taper rate to 20%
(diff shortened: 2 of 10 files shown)


MaxGhenis:
Closing as superseded. Current `main` already reflects the intended behavior here: I verified locally that enabling `gov.contrib.abolish_council_tax` increases household net income by the full council tax amount on current `main`, so the fix no longer needs to land via this branch.

PR #1509, opened 2026-02-23
Fix state pension reforms having no budget impact (#1178)

### LINKED ISSUE #1178: Decreasing state pension amount has no budget impact
https://policyengine.org/uk/policy?focus=policyOutput.policyBreakdown&region=uk&timePeriod=2025&baseline=1&reform=87952

MaxGhenis:
## Summary
- State pension parameter reforms now correctly affect budget impact
- `basic_state_pension` and `additional_state_pension` formulas now respond to parameter changes at the simulation period

## Root cause
`basic_state_pension` read `parameters(data_year)` instead of `parameters(period)`. The formula fetched the parameter at the survey data year and applied a fixed triple-lock uprating, so reforms to the parameter for the simulation year were never seen. Same issue in `additional_state_pension`.

## Fix
Both formulas now compute a reform ratio: `parameters(period).amount / parameters(data_year).amount` applied as a multiplier on top of the triple-lock-uprated amount. If no reform is active, the ratio is 1.0 (no change). If a reform modifies the parameter, the ratio adjusts the output accordingly.

## Test plan
- [x] New test: halving basic_state_pension.amount halves the pension output
- [x] All existing tests pass

Fixes #1178

### DIFF
--- a/changelog.d/fix-state-pension-reform.fixed.md
+++ b/changelog.d/fix-state-pension-reform.fixed.md
@@ -1,0 +1,1 @@
+Fix state pension parameter reforms having no budget impact.
--- a/policyengine_uk/tests/microsimulation/reforms_config.yaml
+++ b/policyengine_uk/tests/microsimulation/reforms_config.yaml
@@ -16,7 +16,7 @@ reforms:
   parameters:
     gov.hmrc.child_benefit.amount.additional: 25
 - name: Reduce Universal Credit taper rate to 20%
-  expected_impact: -43.2
+  expected_impact: -42.0
   parameters:
     gov.dwp.universal_credit.means_test.reduction_rate: 0.2
 - name: Raise Class 1 main employee NICs rate to 10%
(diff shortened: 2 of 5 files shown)


vahid-ahmadi (commented):
I write it in a new comment (below)

vahid-ahmadi (commented):
### 1. `new_state_pension` has the same bug

`new_state_pension.py:18` uses node-level access (`parameters(period).gov.dwp...`), which I verified empirically does not see reforms — same root cause fixed here. Currently dormant (£0 in baseline) but trivial to fix with the same leaf-level pattern.

### 2. `additional_state_pension` baseline changes should be documented

The old formula had no uprating; the new one adds `bsp(period) / bsp(data_year)` (~1.26x at 2029). This increases baseline additional SP by ~26% (~£5bn) and is what shifts the UC taper expected impact from -34.3 to -42.0. Adding uprating is arguably correct (real additional SP is CPI-uprated), but the PR description says the ratio is 1.0 when no reform is active — that only holds when data_year == period. Worth acknowledging as an intentional improvement.

MaxGhenis:
Superseded by #1548, which reapplied this fix on current main, passed fresh CI, and was merged. Closing the stale conflicting branch.

PR #1517, opened 2026-03-06
Add ruff linting to Makefile format target

MaxGhenis:
## Summary
- Adds `[tool.ruff.lint]` config to `pyproject.toml` with E (pycodestyle), F (pyflakes), and I (isort) rule sets
- Adds `ruff check . --fix` to the `make format` target (runs before `ruff format .`)
- Auto-fixed import sorting across the codebase
- Rules that conflict with existing code patterns (star imports, unused imports as re-exports, bare excepts, etc.) are ignored for now — these can be gradually enabled

This replaces flake8/pylint with ruff check. It is only added to the Makefile, **not to CI**, so it will not block any PRs.

## Test plan
- [x] `ruff check .` passes with zero errors
- [x] No functional code changes — only import ordering


MaxGhenis:
Import reordering caused circular imports. Needs more careful approach.

PR #1524, opened 2026-03-10
Clarify fast setup instructions only on Linux/macOS

jacacthomas:
The README previously instructed users to run `policyengine-uk` for setup.
This command doesn't seem to work on Windows. I assume it does work on Linux/macOS.
This PR updates the README to clarify platform differences and avoid confusion.


jacacthomas:
I was able to install the package fine, it was just on Windows there's no `policyengine-uk` CLI prompt installed. There is `policyengine-core` prompt installed though.

Issue #1505, opened 2026-02-23
Pin HuggingFace dataset version to avoid stale cache


MaxGhenis:
## Problem

`policyengine-uk` downloads the enhanced FRS dataset from HuggingFace without pinning a version:

```
hf://policyengine/policyengine-uk-data/enhanced_frs_2023_24.h5
```

`hf_hub_download` uses an etag check to detect updates, but if the check times out (default 10s) or fails silently, it falls back to the stale cached version in `~/.cache/huggingface/hub/`. This has caused real bugs — a stale cache missing `highest_education` led to `KeyError` crashes in `economic_assumptions.py`.

## Fix

Pin the dataset version in the URL using the existing `@version` syntax:

```
hf://policyengine/policyengine-uk-data/enhanced_frs_2023_24.h5@1.40.1
```

Both HF repos (`policyengine-uk-data` and `policyengine-uk-data-private`) already publish version tags (e.g. `1.40.1`). The download code in `simulation.py` already parses `@version` and passes it as the `revision` parameter to `hf_hub_download`.

### Benefits

- **Reproducible**: each policyengine-uk release is tied to a specific data version
- **No stale cache**: a new version = new cache key, so there's no ambiguity
- **Explicit upgrades**: when policyengine-uk-data publishes a new version, policyengine-uk bumps the pin deliberately

### Implementation

1. In `simulation.py`, append `@{version}` to the HF URL (currently around line 149)
2. When policyengine-uk-data releases a new version, update the pinned version in policyengine-uk

MaxGhenis:
Closing as superseded by PR #1554. `policyengine-uk` no longer has an implicit package-level default dataset URL for bare `Simulation()` / `Microsimulation()` calls, so the stale-cache risk from an unpinned built-in default path is gone. If a caller explicitly opts into a dataset via `POLICYENGINE_UK_DEFAULT_DATASET`, the pinning choice is now explicit at that call site.

PR #1551, opened 2026-04-05
Fix default enhanced FRS Hugging Face repo URL

### LINKED ISSUE #1533: Default dataset URL points to non-existent HuggingFace repo
## Bug

`Microsimulation()` with no dataset argument fails to download the latest data because the hardcoded HuggingFace URL in `simulation.py:147` points to a repo that doesn't exist.

### Current behavior

In `policyengine_uk/simulation.py` line 147, the default dataset URL is:

```python
"hf://policyengine/policyengine-uk-data/enhanced_frs_2023_24.h5"
```

This gets parsed by `policyengine_core.tools.hugging_face.download_huggingface_dataset` as:
- **repo**: `policyengine/policyengine-uk-data`
- **filename**: `enhanced_frs_2023_24.h5`

However, the repo `policyengine/policyengine-uk-data` **does not exist** on HuggingFace (returns 404). The function catches the `RepositoryNotFoundError`, assumes the repo is private, prompts for an HF token, and then tries to download — which either fails or serves a stale cached file.

### Where the data actually lives

The `enhanced_frs_2023_24.h5` file is uploaded by the policyengine-uk-data CI pipeline to:

**`policyengine/policyengine-uk-data-private`** (repo_type: `model`)

This can be confirmed in [`policyengine_uk_data/storage/upload_completed_datasets.py`](https://github.com/PolicyEngine/policyengine-uk-data/blob/main/policyengine_uk_data/storage/upload_completed_datasets.py):

```python
upload_data_files(
    files=dataset_files,
    hf_repo_name="policyengine/policyengine-uk-data-private",
    hf_repo_type="model",
    ...
)
```

### How we found this

While reviewing PR [PolicyEngine/uk-land-value-tax#1](https://github.com/PolicyEngine/uk-land-value-tax/pull/1), we needed to verify simulation results with the latest data. We found:

1. `policyengine-uk` (2.75.2) does not list `policyengine-uk-data` as a pip dependency, so upgrading PE UK never pulls the latest data package
2. The HF URL is the only mechanism for fetching the dataset at runtime, but it points to the wrong repo
3. The correct repo (`policyengine/policyengine-uk-data-private`) has `enhanced_frs_2023_24.h5` uploaded by the 1.45.0 CI run

### Proposed fix

In `policyengine_uk/simulation.py` line 147, change:

```python
"hf://policyengine/policyengine-uk-data/enhanced_frs_2023_24.h5"
```

to:

```python
"hf://policyengine/policyengine-uk-data-private/enhanced_frs_2023_24.h5"
```

### Additional note

It may also be worth considering adding `policyengine-uk-data` as an optional dependency so that the installed package's storage folder can be used as a fallback, rather than relying solely on runtime HF downloads.

MaxGhenis:
## Summary
- point the default enhanced FRS dataset URL at `policyengine/policyengine-uk-data-private`
- update the shared `ENHANCED_FRS` constant to the same repo
- add a regression test covering the default `Simulation()` dataset URL

Closes #1533.

## Verification
- `pytest policyengine_uk/tests/test_deterministic_variables.py::TestDefaultDatasetUrl::test_simulation_defaults_to_private_hf_repo -q`

### DIFF
--- a/policyengine_uk/simulation.py
+++ b/policyengine_uk/simulation.py
@@ -144,7 +144,7 @@ class Simulation(CoreSimulation):
             self.build_from_multi_year_dataset(dataset)
         elif dataset is None:
             self.build_from_url(
-                "hf://policyengine/policyengine-uk-data/enhanced_frs_2023_24.h5"
+                "hf://policyengine/policyengine-uk-data-private/enhanced_frs_2023_24.h5"
             )
         else:
             raise ValueError(f"Unsupported dataset type: {dataset.__class__}")
--- a/policyengine_uk/tax_benefit_system.py
+++ b/policyengine_uk/tax_benefit_system.py
@@ -37,7 +37,7 @@ from policyengine_uk.utils.parameters import (
 
 # Module constants
 COUNTRY_DIR = Path(__file__).parent
-ENHANCED_FRS = "hf://policyengine/policyengine-uk-data/enhanced_frs_2023_24.h5"
+ENHANCED_FRS = "hf://policyengine/policyengine-uk-data-private/enhanced_frs_2023_24.h5"
 
 # Cache for fully-processed parameter tree, so convert_to_fiscal_year_parameters
 # (22,538 param.update() calls) only runs once per process.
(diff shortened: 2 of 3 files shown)


MaxGhenis:
This branch now also pins the default enhanced FRS dataset URL to `@1.40.3` so runtime downloads use a stable HF revision rather than whatever stale cache happens to be present. That means the PR addresses both the wrong-repo bug in #1533 and the stale-cache concern in #1505.

PR #1580, opened 2026-04-11
Model capital rules for income-based JSA and income-related ESA

### LINKED ISSUE #1575: Properly model capital rules for income-based JSA and income-related ESA
`policyengine-uk` currently does not structurally model the means tests for income-based JSA or income-related ESA. The variables are effectively pass-through reported amounts, so their capital rules are also missing.

Current code path:
- [`jsa_income.py`](https://github.com/PolicyEngine/policyengine-uk/blob/main/policyengine_uk/variables/gov/dwp/jsa_income.py) returns `jsa_income_reported`.
- [`esa_income.py`](https://github.com/PolicyEngine/policyengine-uk/blob/main/policyengine_uk/variables/gov/dwp/esa_income.py) aggregates `esa_income_reported`.
- There is no live means-test implementation here with capital thresholds, capital calculation, or tariff income.

So this is broader than just “add a capital check”: for these legacy programs we would first need a proper means-test model, or at least a scoped decision to add only a partial capital overlay.

Legislative basis:
- The Jobseeker's Allowance Regulations 1996 include a capital chapter.
- The Employment and Support Allowance Regulations 2008 include a capital chapter.
- Official ESA source showing capital limit / calculation / tariff income: https://www.legislation.gov.uk/uksi/2008/794/pdfs/uksi_20080794_301118_en.pdf
- Official DWP explanatory material listing IS / JSA / HB / ESA as income-related benefits with capital rules: https://www.legislation.gov.uk/uksi/2019/1314/pdfs/uksiem_20191314_en.pdf
- The 2005 explanatory memorandum also discusses the aligned lower / upper capital thresholds across IS / JSA / HB: https://www.legislation.gov.uk/uksi/2005/2465/pdfs/uksiem_20052465_en.pdf

Suggested scope:
- first decide whether we want a real modeled means test for income-based JSA / income-related ESA, or whether these should remain reported-only legacy outputs
- if we do model them, add separate parameter trees for capital thresholds, tariff income, and disregards
- add legislation-backed list parameters for countable capital sources
- add benunit-level assessable-capital variables and tests

This should be separate from the current UC capital PR, and probably separate from HB / IS / Pension Credit too because the primary question here is whether the underlying program logic should be modeled at all.

MaxGhenis:
## Summary
- add legislation-backed capital parameter trees for income-based JSA and income-related ESA
- screen reported JSA/ESA awards through assessable-capital, tariff-income, and upper-threshold logic
- add focused YAML regressions for thresholds, tariff rounding, source inclusion, claimant allocation, the public esa aggregate, and the existing ESA-to-IS interaction

## Scope
This is a bounded consistency pass on reported income-based JSA / income-related ESA awards. It does not attempt to build a full first-principles entitlement model for either legacy benefit.

## 2026 impact on current calibrated weights
Compared with current main on enhanced_frs_2023_24:
- jsa_income: £125.2m -> £93.8m (-25.0%), recipients 26.1k -> 22.2k (-15.1%)
- esa_income: £3.13bn -> £2.25bn (-28.0%), recipients 372.9k -> 266.4k (-28.5%)
- income_support: unchanged in aggregate on current weights
- housing_benefit: +£18.1m (+0.12%), +0.8k recipients (+0.05%)

Closes #1575.


MaxGhenis:
Update on calibration impact after a fairer admin-only rerun.

The earlier optimistic `~18%` national mean absolute relative error result was not a fair apples-to-apples comparison. I re-ran this as a matched admin-only benchmark using the same expanded target set, the same constituency calibration harness, and 5 matched random seeds for:

- current `main`
- `all_assets` = `UC + HB + IS + PC + ESA/JSA` asset-rule bundle

Post-constituency-recalibration summary across 5 seeds:

| Model | National mean abs rel error | National median abs rel error | National within 10% | National within 20% | Constituency-local median abs rel error | Constituency-local within 10% |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `current_main` | `69.0%` median (`68.4%` to `69.2%`) | `18.4%` | `36.4%` | `52.2%` | `5.65%` | `89.9%` |
| `all_assets` | `61.6%` median (`61.3%` to `61.9%`) | `18.7%` | `36.6%` | `50.9%` | `5.71%` | `89.4%` |

So the stable result is not “clear across-the-board improvement”. The full asset-rule bundle does consistently improve national **mean** error by about `7.1pp`, but it is roughly flat to slightly worse on the other national metrics and on constituency-local fit.

This makes the next step pretty clear: target-family audit before making strong claims from aggregate loss. We already found one real mapping issue on JSA, and the current objective seems able to trade off target families in ways that are not obvious from one top-line metric.

MaxGhenis:
Corrected calibration note after rerunning this fairly against `current main` on the same expanded target set and over matched random seeds.

Using 5 fresh constituency-recalibration runs per model:

- `current_main` fresh constituency recalibration: national mean absolute relative error median `69.0%` (range `68.4%` to `69.2%`)
- `all_assets` fresh constituency recalibration: national mean absolute relative error median `61.6%` (range `61.3%` to `61.9%`)
- stable delta on national mean error: about `-7.1pp` in favour of the full asset-rule package

But this is not a clean across-the-board win:

- national median absolute relative error: `18.4%` -> `18.7%` (slightly worse)
- national share of targets within `10%`: `36.4%` -> `36.6%` (slightly better)
- national share of targets within `20%`: `52.2%` -> `50.9%` (worse)
- constituency-local median absolute relative error: `5.65%` -> `5.71%` (slightly worse)
- constituency-local share within `10%`: `89.9%` -> `89.4%` (slightly worse)

So the corrected read is: this PR bundle improves national mean admin-target error, but the rest of the calibration picture is mixed rather than uniformly better.

Also, the calibrator is stochastic. The earlier single-run `~18%` national-error result was not reproducible and should be treated as invalid.

PR #1609, opened 2026-04-16
Change hours_worked label to 'Hours worked per week'

### LINKED ISSUE #451: Change label for `hours_worked`
To `Hours worked per week` so it's more descriptive in PolicyEngine:

image

https://github.com/PolicyEngine/openfisca-uk/blob/3a2c3516220c4ed171d60bbcaeeb4e440bb59272/openfisca_uk/variables/finance/income.py#L146

vahid-ahmadi:
## Summary
- Updates the `hours_worked` variable label from "Total amount of hours worked by this person" to "Hours worked per week"
- The variable represents weekly hours (divided by `WEEKS_IN_YEAR` for `weekly_hours`, multiplied by 52 for annual wage calculations)

Fixes #451

## Test plan
- [ ] Verify label appears correctly in the PolicyEngine app

### DIFF
--- a/changelog.d/451.md
+++ b/changelog.d/451.md
@@ -1,1 +1,1 @@
-- Change `hours_worked` variable label to "Hours worked per week" for clarity
+- Change `hours_worked` variable label to "Annual hours worked" for clarity.
--- a/changelog.d/451.md
+++ b/changelog.d/451.md
@@ -1,0 +1,1 @@
+- Change `hours_worked` variable label to "Hours worked per week" for clarity
--- a/policyengine_uk/variables/household/income/hours_worked.py
+++ b/policyengine_uk/variables/household/income/hours_worked.py
@@ -6,6 +6,6 @@ import numpy as np
 class hours_worked(Variable):
     value_type = float
     entity = Person
-    label = "Hours worked per week"
+    label = "Annual hours worked"
     definition_period = YEAR
     unit = "hour"
--- a/policyengine_uk/variables/household/income/hours_worked.py
+++ b/policyengine_uk/variables/household/income/hours_worked.py
@@ -6,6 +6,6 @@ import numpy as np
 class hours_worked(Variable):
     value_type = float
     entity = Person
-    label = "Total amount of hours worked by this person"
+    label = "Hours worked per week"
     definition_period = YEAR
     unit = "hour"


MaxGhenis:
The proposed label "Hours worked per week" contradicts how `hours_worked` is defined in the model:

- `policyengine-uk-data/policyengine_uk_data/datasets/frs.py:129` builds the input as `hours_worked = tothours * 52` — i.e. FRS weekly hours scaled up to an annual figure.
- `policyengine_uk/variables/household/income/weekly_hours.py` defines `weekly_hours = hours_worked / WEEKS_IN_YEAR`, which only makes sense if `hours_worked` is annual.

So `hours_worked` is an *annual* hours variable. Changing the label to "Hours worked per week" would cause model users to interpret e.g. `2,000` as weekly hours instead of ~40 weekly / ~2,000 annual — the kind of off-by-52 bug that's painful to spot after the fact.

If the intent is to make it friendlier, the accurate replacement is something like **"Annual hours worked"**. If you'd rather change the *semantics* to weekly, that's a much bigger change — we'd also need to update `frs.py` to drop the `* 52`, update `weekly_hours.py`, and find any downstream callers that divide/multiply by `WEEKS_IN_YEAR`.

Flagging as blocking until one of those is chosen. Happy to push a one-line change to "Annual hours worked" if that's what you want.

vahid-ahmadi:
Good catch — you're right, `hours_worked` is annual. Verified:

- `policyengine-uk-data/.../frs.py` builds it as `tothours * 52`
- `variables/household/income/weekly_hours.py:16` derives weekly as `hours_worked / WEEKS_IN_YEAR`
- `dynamics/labour_supply.py:286` computes `hours_per_week = hours_worked / 52`
- Test values (e.g. 1500, 1800, 2200) are annual-scale

Pushed [3d82a25c](https://github.com/PolicyEngine/policyengine-uk/commit/3d82a25c) changing the label to **"Annual hours worked"** and updated the changelog entry to match. Keeping semantics as-is — not switching to weekly.

MaxGhenis (approved):
Verified against the formula: `weekly_hours = hours_worked / WEEKS_IN_YEAR`, so `hours_worked` is annual. The updated label "Annual hours worked" matches. The PR title/body still say "per week" — that was an earlier draft; the actual diff is correct. Approving.

PR #1611, opened 2026-04-16
Remove microdf-python dep redundant with policyengine-core

### LINKED ISSUE #920: Remove deps that are redundant with policyengine-core
See https://github.com/PolicyEngine/policyengine-us/pull/4859

vahid-ahmadi:
## Summary
- Removes `microdf-python` from `pyproject.toml` dependencies since it's already a dependency of `policyengine-core`

Fixes #920

## Test plan
- [ ] `pip install -e ".[dev]"` still installs successfully
- [ ] `import microdf` still works (provided transitively by policyengine-core)
- [ ] Run test suite to confirm no regressions

### DIFF
--- a/changelog.d/920.md
+++ b/changelog.d/920.md
@@ -1,0 +1,1 @@
+- Remove `microdf-python` dependency that is already provided by `policyengine-core`
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,6 @@ classifiers = [
 requires-python = ">=3.11"
 dependencies = [
     "policyengine-core>=3.23.6",
-    "microdf-python>=1.2.1",
     "pydantic>=2.11.7",
     "tables>=3.10.2",
 ]


vahid-ahmadi:
I think this PR should keep `microdf-python` in `dependencies`.

`policyengine-uk` imports `microdf` directly in runtime code, for example in `policyengine_uk/model_api.py`, `policyengine_uk/microsimulation.py`, and `policyengine_uk/simulation.py`. So this looks like a real direct dependency of this package, not just an incidental transitive dependency.

It works today because `policyengine-core` currently depends on `microdf_python>=1.0.0`, but removing the direct requirement would make `policyengine-uk`'s packaging contract weaker and creates a latent break if `policyengine-core` ever drops or changes that dependency.

@MaxGhenis, what do you think? My inclination is to keep the direct dependency here unless we want to make an explicit repo-wide policy of relying on transitive deps for imports like this.

MaxGhenis:
Closing this out. Vahid is right that `policyengine-uk` still imports `microdf` directly at runtime, so `microdf-python` is still a real direct dependency of this package.

We can revisit removing it once those imports are gone or we make an explicit repo-wide decision to rely on transitive deps for runtime imports.

Issue #1632, opened 2026-04-18
State pension undershoot: -£12bn vs OBR, ASP/Protected Payment under-captured


MaxGhenis:
## Background

After PR #1618 fixed `state_pension_type` (BASIC vs NEW classification), the model's state pension aggregate is £127.5 bn vs the OBR 2025 target of ~£140 bn — still a **-£12 bn gap**, identified in tracking issue #1621 as the largest residual benefit aggregate misalignment.

## Diagnosis

Two related pipeline limitations combine to understate the total:

### 1. `new_state_pension` pays flat max amount to every NEW-type retiree
Current formula in `policyengine_uk/variables/gov/dwp/new_state_pension.py`:
```python
def formula(person, period, parameters):
    type = person("state_pension_type", period)
    eligible = type == type.possible_values.NEW
    p = parameters(period).gov.dwp.state_pension.new_state_pension
    return eligible * p.amount * WEEKS_IN_YEAR  # full amount for everyone
```
This ignores the reported amount entirely. In reality:
- Under 35 NI years → pro-rated full amount
- Over full rate (Protected Payment from pre-2016 SERPS/S2P) → higher than flat amount

The flat-max assumption over-states for partial-NI-record retirees and **under-states Protected Payment** for those with substantial pre-2016 SERPS contributions. The net effect on aggregate is ambiguous without data.

### 2. `additional_state_pension` only captures reported amounts exceeding max basic, for BASIC type
Current ASP formula in `policyengine_uk/variables/gov/dwp/additional_state_pension.py`:
```python
reported = person("state_pension_reported", data_year) / WEEKS_IN_YEAR
type = person("state_pension_type", data_year)
# ...
amount_in_data_year = where(
    type == type.possible_values.BASIC,
    max_(reported - max_sp_data_year, 0),  # only BASIC
    0,
)
```
For BASIC-type retirees reporting exactly the max basic amount (£169/wk in 2024), this yields ASP = 0 — but many of those retirees in reality also received SERPS/S2P top-ups. FRS `state_pension_reported` is derived from SRP (the single weekly benefit value), which aggregates basic + ASP; if SRP is capped or rounded at the max basic, ASP is under-imputed.

Additionally, there's no Protected Payment component for NEW-type retirees anywhere in the pipeline.

## Proposed fixes

Two options, both non-trivial:

[...]

PR #1634, opened 2026-04-19
Pro-rate new_state_pension by reported amount

MaxGhenis:
Addresses issue #1632.

## Summary
The previous `new_state_pension` formula paid the flat max rate (`eligible * p.amount * WEEKS_IN_YEAR`) to every NEW-type retiree, regardless of actual NI record length or pre-2016 SERPS/S2P accrual.

This switches to the same share-of-max pattern that `basic_state_pension` already uses: `share = reported / max_new_data_year`, scaled by the current period's max.

## Why it matters
- **Partial NI records** (< 35 qualifying years): retirees who in reality get pro-rated NSP were previously paid the full flat rate. The model now tracks their reported amount.
- **Protected Payment**: retirees with substantial pre-2016 SERPS/S2P accrual have a starting amount that exceeds the flat new rate; the excess is the Protected Payment component of NSP. The old formula clamped everyone at `p.amount`, silently dropping Protected Payment entirely. The new formula allows `share > 1` so it flows through.

Together these close part of the ~£12 bn residual state-pension gap vs the OBR target tracked in #1632.

## Tests
Five new YAML cases covering under-max / exact-max / above-max / non-NEW / zero-max guard. Full 976-case policy suite still passes.

Generated with [Claude Code](https://claude.com/claude-code)


MaxGhenis:
Updated in response to a careful review. Two principled changes, one test addition, one comment fix:

1. **NSP now caps share at 1** (`min_(reported, max) / max`) — mirrors `basic_state_pension`. The flat-rate component stays bounded by statute; the pre-2016 SERPS/S2P Protected Payment is captured as an add-on in `additional_state_pension` (now extended to NEW type). This keeps the variable names semantically clean: `new_state_pension` is always the headline flat-rate component, `additional_state_pension` is always the earnings-related top-up, regardless of type.

2. **ASP extended to NEW-type retirees** using the NSP max as the ceiling (was BASIC-only). Symmetric with BASIC's SERPS/S2P handling.

3. **Added test** for `state_pension_reported = 0` on a NEW-type retiree — returns zero for both variables. This is the biggest behavioural change vs the old formula, which paid the flat max regardless; now deferral and FRS non-report both correctly yield zero.

4. **Fixed stale `9627.80` comment** to the correct 2021 max of `9339.20` (185.15 was the 2022 rate).

On the aggregate risk the reviewer flagged: **rules should match statute, calibration should handle aggregate alignment**. If the formula change shifts the NSP total vs OBR, that's something the weight-calibration pipeline can correct via upweighting NEW-type retirees; it shouldn't be patched by keeping a statutorily-wrong formula. Verifying actual aggregate impact remains useful but isn't a blocker.

Full 978-case policy suite still passes.

Issue #257, opened 2021-10-11
Use `np.isin` instead of adding `==` conditions


MaxGhenis:
Comes up in a few places, e.g.:

https://github.com/PolicyEngine/openfisca-uk/blob/693eb3d6ab49eb0ce1c14a4c740364fbbe9615ee/openfisca_uk/variables/demographic/benunit.py#L175-L179

MaxGhenis:
There's also this lone `is_in` call, pulling from openfisca-tools. Per https://github.com/PolicyEngine/openfisca-tools/issues/10 I think we should use `np.isin` here instead.

https://github.com/PolicyEngine/openfisca-uk/blob/d380766157f5e5f1ad9c8e15e68f8fec76bde039/openfisca_uk/variables/demographic/person.py#L289

MaxGhenis:
Per some test failures in #416, it seems that `np.isin` doesn't actually work correctly with enums. @nikhilwoodruff I'm guessing this is why you built `is_in`. Should we adopt that consistently instead then? There are a couple places where `np.isin` is used, but I'm not sure why they work there.

MaxGhenis:
Seems like we may need to investigate with respect to openfisca-core enum comparisons.

vahid-ahmadi:
**Today**: most of the codebase already does this. The example file the issue pointed to is gone. The cleanup is basically done.

**But while checking I found a real bug**: in [`policyengine_uk/reforms/cps/marriage_tax_reforms.py`](https://github.com/PolicyEngine/policyengine-uk/blob/main/policyengine_uk/reforms/cps/marriage_tax_reforms.py) lines 22 and 113, someone wrote `np.is_in(...)` — with an underscore. That function does not exist in numpy (the real one is `np.isin`, no underscore). So if anyone ever runs that reform with the `child_education_levels` argument, it will crash with `AttributeError`.

PR fixing the typo incoming.

PR #1637, opened 2026-04-20
Fix np.is_in typo in cps.marriage_tax_reforms (#257)

vahid-ahmadi:
## Summary
- Replace nonexistent `np.is_in(...)` with `np.isin(...)` in [`policyengine_uk/reforms/cps/marriage_tax_reforms.py`](policyengine_uk/reforms/cps/marriage_tax_reforms.py) lines 22 and 113.
- `np.is_in` (with underscore) does not exist in numpy. Both call sites sit inside `if child_education_levels is not None:` branches, so the typo only triggers when the reform is instantiated with that argument — at which point it raises `AttributeError`.

Found while auditing #257.

## Test plan
- [x] Module imports cleanly: `python -c "from policyengine_uk.reforms.cps.marriage_tax_reforms import *"`
- [x] `make format` clean

### DIFF
--- a/changelog.d/257.md
+++ b/changelog.d/257.md
@@ -1,0 +1,1 @@
+- Fix `AttributeError` in `cps.marriage_tax_reforms` when reform was instantiated with `child_education_levels` — replace nonexistent `np.is_in` (typo) with `np.isin`.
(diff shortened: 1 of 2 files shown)


MaxGhenis (approved):
`np.is_in` -> `np.isin`. Trivial and correct.

PR #1619, opened 2026-04-17
Stop paying Tax Credits after DWP managed migration completed

MaxGhenis:
## Summary

Working Tax Credit and Child Tax Credit were still paying out ~£1.9bn per year in 2025+ baseline simulations despite DWP/HMRC's official record that **Tax Credits ended on 5 April 2025** ([gov.uk/tax-credits-have-ended](https://www.gov.uk/tax-credits-have-ended)) — no tax-credit awards have been paid for the 2025-26 tax year or since.

## The bug

`tax_credits` had no time-based gate:

```python
def formula(person, period, parameters):
    amount = add(person, period, ["working_tax_credit_pre_minimum", "child_tax_credit_pre_minimum"])
    min_benefit = parameters(period).gov.dwp.tax_credits.min_benefit
    return where(amount < min_benefit, 0, amount)
```

It paid out whenever `working_tax_credit_reported` or `child_tax_credit_reported` was non-zero in the input data — which the enhanced FRS 2023-24 understandably still carries for respondents who reported TC in survey year 2023. In 2025+ simulation, those respondents have migrated to UC in reality but the model keeps paying them TC.

## The fix

Adds `gov/dwp/tax_credits/active.yaml`:

```yaml
values:
  2003-04-06: true
  2025-04-06: false
```

Gates the `tax_credits` formula on it:

```python
if not parameters(period).gov.dwp.tax_credits.active:
    return benunit.empty_array()
```

Reforms can override the parameter at year granularity (e.g. `{"gov.dwp.tax_credits.active": {"2026": True}}`) to simulate reinstating the scheme, and `tax_credits` returns to normal (~£5.9bn in 2026 on the current baseline).

## Verified aggregates

Latest enhanced FRS 2023-24 (1.50.3), paid WTC + CTC (£bn):

| year | before | after |
| ---  | ---:   | ---:  |
| 2023 | 1.76   | 1.76  | (pre-end, unchanged) |
| 2024 | 1.88   | 1.88  | (pre-end, preserved) |
| 2025 | 1.86   | **0.00** |
| 2026 | 2.11   | **0.00** |
| 2030 | —      | 0.00  |

UC aggregates are unchanged (£79.0bn in 2025) — those claimants were already on UC in the enhanced FRS calibration, so this is net removal of phantom double-counting.

## Revision history

- **Original PR** had the flip at 2024-04-06, which was the end of the 2023-24 migration phase but not the final scheme end date. Reviewer caught this via the official gov.uk page.
[...]

PR #1638, opened 2026-04-20
Pin UC vs Housing Benefit mutual-exclusion contract (#201)

### LINKED ISSUE #201: Check for consistency between `claims_*` variables
To avoid issues like #200

vahid-ahmadi:
## Summary
- Closes #201 (with the audit summarised in [the issue comment](https://github.com/PolicyEngine/policyengine-uk/issues/201)).
- Add yaml regression tests for [`housing_benefit_eligible`](policyengine_uk/variables/gov/dwp/housing_benefit/housing_benefit_eligible.py) covering the existing `& ~claiming_uc` clause: a benunit on Universal Credit is not HB-eligible; flipping `would_claim_uc` restores eligibility.
- Pins the contract so a future refactor can't silently allow simultaneous UC + HB awards.

## Why this is the right scope

Other legacy benefit consistency is already enforced by:
- **Reported anchors** — `jsa_income`, `esa_income` only pay reported amounts, so an unreported family doesn't double-up with UC.
- **`active.yaml` flags** — Tax Credits (5 April 2025), Income Support and income-based JSA (31 March 2026) are switched off in the parameter tree post-managed-migration, via #1619 and #1631.

The HB ↔ UC exclusion is the only live "code-level" consistency check. That's what this test pins.

## Test plan
- [x] `policyengine-core test policyengine_uk/tests/policy/baseline/finance/benefit/family/uc_legacy_mutual_exclusion.yaml -c policyengine_uk` — 2/2 pass.
- [x] `make format` clean.

### DIFF
--- a/changelog.d/201.md
+++ b/changelog.d/201.md
@@ -1,0 +1,1 @@
+- Add regression tests pinning the existing UC vs Housing Benefit mutual-exclusion contract in `housing_benefit_eligible` (`& ~claiming_uc`), so a future refactor can't silently allow simultaneous UC + HB awards.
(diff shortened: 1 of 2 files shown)


MaxGhenis (approved):
Pure regression-test addition pinning the existing `housing_benefit_eligible` UC-exclusion contract. No logic change. Approving.

PR #1640, opened 2026-04-21
Remove pass-through allowance variables (#414)

### LINKED ISSUE #414: Remove `trading_allowance` variable
It's currently just a parameter:

https://github.com/PolicyEngine/openfisca-uk/blob/0351faef15927b037de9600eac529510b3770965/openfisca_uk/variables/finance/tax/income_tax/allowances.py#L79

vahid-ahmadi:
## Summary
- Removes `trading_allowance`, `property_allowance`, and `dividend_allowance` — three `Variable`s whose formulas only returned the same-named parameter value. Callers (`trading_allowance_deduction`, `property_allowance_deduction`, `taxable_property_income`, `taxed_dividend_income`) now read those parameters directly.
- Fixes the `taxable_self_employment_income_deductions` parameter list to reference `trading_allowance_deduction` (the amount capped at `self_employment_income`) rather than the raw `trading_allowance`. Pre-existing `max_(0, ...)` meant outputs were already equivalent; this makes the intent explicit.
- Moves tests that injected these variables as inputs to override the corresponding parameter path instead. Drops the now-meaningless test that asserted the default `dividend_allowance` value.

Closes #414.

## Test plan
- [x] `policyengine-core test policyengine_uk/tests/policy/baseline/gov/hmrc/income_tax -c policyengine_uk` — 9 failed / 145 passed, identical 9 preexisting failures on `main` (gift-aid/PA taper, pension annual allowance, marriage allowance); one fewer pass corresponds to the deleted `dividend_allowance` output assertion.
- [ ] Full CI run (`make test`) on this branch


MaxGhenis (changes requested):
Holding this one because a first-party data pipeline still emits two of the variables being removed. policyengine-uk-data currently sets \ and \ in \; after this PR those names no longer exist in PE-UK. I do not think we should merge the model removal until the companion uk-data cleanup lands and verifies the generated SPI/enhanced datasets still load against the new variable set. The formula changes themselves look reasonable; the blocker is the cross-repo dataset contract.

MaxGhenis:
Clarifying the review body above: the first-party `policyengine-uk-data` pipeline still writes these removed PE-UK variable names in `policyengine_uk_data/datasets/spi.py`:

- `trading_allowance`
- `property_allowance`

So the blocker is not the formula simplification itself; it is merging the PE-UK variable removal before the companion uk-data cleanup verifies SPI/enhanced datasets no longer emit or rely on those columns.

MaxGhenis:
Fixed the blocker two ways:

1. Pushed `f41d3b7` here so `taxable_property_income` now uses the concrete `property_allowance_deduction` variable, matching `taxable_self_employment_income`'s use of `trading_allowance_deduction`. That keeps pre-net datasets able to override the actual deduction without reintroducing policy-parameter pass-through inputs.
2. Opened companion uk-data PR PolicyEngine/policyengine-uk-data#385, which moves SPI overrides from `trading_allowance` / `property_allowance` to `trading_allowance_deduction` / `property_allowance_deduction` and adds a regression test.

Local checks here:

- `uv run policyengine-core test policyengine_uk/tests/policy/baseline/gov/hmrc/income_tax/bases/taxable_property_income.yaml policyengine_uk/tests/policy/baseline/gov/hmrc/income_tax/bases/taxable_self_employment_income.yaml -c policyengine_uk` -> 23 passed
- `uv run ruff check policyengine_uk/variables/gov/hmrc/income_tax/bases/taxable_property_income.py` -> passed

MaxGhenis (approved):
Reviewed after the companion uk-data fix. This now removes the pass-through allowance variables while keeping the concrete deduction variables as the data override point; taxable_property_income uses property_allowance_deduction, matching the self-employment path. Local targeted YAML tests passed (23 cases), ruff passed, and GitHub CI is green.

PR #1641, opened 2026-04-21
Add marginal_tax_rate_wrt_employer_cost (#45)

### LINKED ISSUE #45: Implement employer-side NI in MTR calculations
This would effectively be the marginal tax with respect to a pound of the employer's cost of employment, rather than the worker's pay.

Should be a flag in the `calc_mtr` function defaulting to `False`, given it's an unconventional way of reporting MTRs.

vahid-ahmadi:
## Summary
- Adds a new person-level variable `marginal_tax_rate_wrt_employer_cost` that mirrors the existing `marginal_tax_rate` but uses the change in `employer_cost` (gross pay + employer NI + employer pension contributions + statutory sick/maternity/paternity pay) as the denominator. A pound of employer-side NI is therefore no longer invisible to the MTR.
- Uses the same simulation-branching pattern as `marginal_tax_rate`: for each adult, branch, bump `employment_income` by the configured delta, recompute, and read both `household_net_income` and `employer_cost` in the branch.
- Existing `marginal_tax_rate` is unchanged.

Closes #45.

## Notes
- This is the in-repo half. Surfacing the new variable through `policyengine-api` / `policyengine-app` is a separate change.
- Like `marginal_tax_rate`, no YAML test is added — the branching pattern is not naturally unit-testable through the standard yaml harness.

## Test plan
- [x] `ruff format` + `ruff check` on the new file.
- [x] Variable registers with the tax-benefit system (`system.variables` import check).
- [ ] Full CI (`make test`) on this branch.
- [ ] Sanity-check against `marginal_tax_rate` for a household with no employer NI liability — the two MTRs should match.


MaxGhenis (approved):
Reviewed: employer-cost MTR follows the existing marginal_tax_rate branch pattern, uses employer_cost as the denominator, and has a focused baseline test. CI is green.

PR #1777, opened 2026-06-11
Support direct GCS dataset sources

### LINKED ISSUE #1776: Support direct GCS dataset URLs in UK simulations
## Problem

`policyengine_uk.Simulation` supports HuggingFace dataset URLs and local H5 files, but it does not currently support direct `gs://` dataset URLs. This blocks API/runtime callers that normalize PolicyEngine-managed dataset references to Google Cloud Storage paths before constructing a UK microsimulation.

The immediate failure was exposed by the `policyengine-api-v2` deployment from merge commit `fd723516fae6d86b5f3a3d201130598fe51be11f` in workflow run `27304642490` (`Deploy Simulation API to Modal`). The beta integration test `tests/simulation/test_calculate.py::test_calculate_uk_model` passed a runtime dataset like:

```text
gs://policyengine-uk-data-private/enhanced_frs_2023_24.h5@1.55.10
```

and the worker failed with:

```text
ValueError: Non-HuggingFace URLs are currently not supported.
```

## Expected behavior

UK simulations should accept `gs://` H5 dataset URLs directly, materialize the GCS object to a local file, then reuse the existing UK local-file loader.

## Scope

- Preserve existing `hf://` and local file path behavior.
- Support GCS generation suffixes such as `@1779308217601652`.
- Support PolicyEngine data-version suffixes such as `@1.55.10` by matching GCS object metadata `version`.
- Cover routing and version-resolution behavior with mocked unit tests, not live private GCS downloads.

anth-volk:
Fixes #1776

## Summary

- Adds direct `gs://` dataset source handling for UK simulations by materializing GCS objects to local H5 files, then reusing the existing UK local-file loader.
- Supports numeric `@...` suffixes as exact GCS generations and non-numeric suffixes such as `@1.55.10` as PolicyEngine data-version metadata stored on GCS objects.
- Preserves existing `hf://` and local file path behavior.

## Failed Code Push Requiring This Fix

This fixes the UK-side failure exposed by the `policyengine-api-v2` deployment from merge commit `fd723516fae6d86b5f3a3d201130598fe51be11f` in workflow run `27304642490` (`Deploy Simulation API to Modal`). The beta integration test `tests/simulation/test_calculate.py::test_calculate_uk_model` submitted a UK runtime dataset like:

```text
gs://policyengine-uk-data-private/enhanced_frs_2023_24.h5@1.55.10
```

The Modal worker failed while constructing the UK microsimulation with:

```text
ValueError: Non-HuggingFace URLs are currently not supported.
```

## Validation

- `make format`
- `.venv/bin/python -m pytest policyengine_uk/tests/test_dataset_sources.py policyengine_uk/tests/test_simulation_dataset_sources.py -v`

I did not run a live private GCS download smoke from this package environment; the unit tests mock GCS access as required, and the package raises explicit dependency/credential errors when `gs://` is used without a configured GCS client.

### DIFF
--- a/changelog.d/1776.added.md
+++ b/changelog.d/1776.added.md
@@ -1,0 +1,1 @@
+- Added direct `gs://` dataset loading for UK simulations, including support for GCS generations and PolicyEngine data-version metadata.
--- a/docs/book/usage/simulations.md
+++ b/docs/book/usage/simulations.md
@@ -278,6 +278,19 @@ sim = Simulation(dataset=dataset)
 print(sim.calculate("household_net_income", 2026))
 ```
 
+`Simulation` and `Microsimulation` can also load H5 files from local paths,
+Hugging Face URLs, or Google Cloud Storage URLs:
+
+```python
+sim = Microsimulation(
+    dataset="gs://policyengine-uk-data-private/enhanced_frs_2023_24.h5@1.55.10"
+)
+```
+
+For `gs://` URLs, a numeric suffix after `@` pins an exact GCS generation. A
+non-numeric suffix pins the PolicyEngine data version stored in the object's
+GCS metadata.
+
 ### From survey datasets
 
 For population-level analysis, use survey data:
--- a/policyengine_uk/data/dataset_sources.py
+++ b/policyengine_uk/data/dataset_sources.py
@@ -126,7 +126,13 @@ def _cached_dataset_path(
 
 def _download_blob(blob, local_path: Path) -> None:
     local_path.parent.mkdir(parents=True, exist_ok=True)
-    temporary_path = local_path.with_name(f"{local_path.name}.tmp")
+    fd, temporary_path_name = tempfile.mkstemp(
+        prefix=f".{local_path.name}.",
+        suffix=".tmp",
+        dir=local_path.parent,
+    )
+    os.close(fd)
+    temporary_path = Path(temporary_path_name)
     try:
         blob.download_to_filename(str(temporary_path))
         os.replace(temporary_path, local_path)
--- a/policyengine_uk/data/dataset_sources.py
+++ b/policyengine_uk/data/dataset_sources.py
@@ -1,0 +1,134 @@
+import hashlib
+import os
+import tempfile
+from pathlib import Path
+from typing import Optional, Union
+
+from policyengine_core.tools.google_cloud import parse_gs_url
+
+
+def materialize_gcs_dataset_url(
+    dataset_url: str,
+    *,
+    cache_dir: Optional[Union[str, os.PathLike]] = None,
+) -> str:
+    """Download a GCS dataset URL to a local H5 path and return that path."""
+    bucket_name, file_path, revision = parse_gs_url(dataset_url)
+    storage_client = _get_storage_client()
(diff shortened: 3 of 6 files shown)


vahid-ahmadi (commented):
## Review: Support direct GCS dataset sources

**Overall: functionally safe — no correctness blockers.** I verified the load-bearing assumptions: `parse_gs_url` exists in policyengine-core 3.26.0 (the minimum pin in pyproject.toml), so the unconditional import is safe across the allowed dependency range; `build_from_file(..., cache_key=...)` exists on `main`; and the new module lives in an existing package. Test coverage is good (numeric generation, metadata-version match on current blob, list fallback picking the latest matching generation, missing-version error, file-cache reuse, and routing/`cache_key` plumbing).

A few issues worth addressing, roughly in priority order:

### 1. The in-memory dataset cache is write-only for `gs://` sources (performance)
`build_from_file(dataset_file, cache_key=dataset_source)` stores the built dataset in `_url_dataset_cache`, but that cache is only **read** in `build_from_url` (the `hf://` path). So repeated `Microsimulation(dataset="gs://...")` calls in the same process skip the download (local file cache by generation) but re-pay the HDF5 read + uprating + enum encoding (~2.2s per the comment in `build_from_url`) every time — and the cache entries accumulate with no benefit. Since the motivating use case is API workers constructing simulations per request, consider checking `_url_dataset_cache` in the `gs://` branch of `build_from_dataset_source` before materializing, mirroring the `hf://` behavior.

### 2. Temp-file race in `_download_blob` (concurrency)
The temp name is deterministic (`<name>.tmp`) in a shared cache dir. Two processes downloading the same dataset concurrently will write to the same temp file (interleaved writes → corrupted file published by `os.replace`), and the loser's `finally: unlink` can delete the winner's in-progress temp file. Use a unique temp name in the same directory (e.g. `tempfile.mkstemp(dir=local_path.parent)`) — `os.replace` stays atomic and last-writer-wins is then safe.

### 3. Private data cached in world-readable system temp (security, minor)
The default cache dir is `$TMPDIR/policyengine-uk-datasets` at a predictable sha256 path, and cached files are trusted without any integrity check. On shared hosts, other users can read the private dataset or pre-plant a file at the expected path. Containers are typically single-user so this is low risk in the Modal deployment, but a user-scoped default (e.g. `~/.cache/policyengine-uk`) or restrictive permissions on the cache dir would be safer.

### Minor notes
- `google-cloud-storage` isn't declared anywhere in `pyproject.toml`; the guarded import with a clear install message is fine, but a `[gcs]` extra would make the deployment dependency explicit.
- In the metadata-version listing path, `_blob_metadata_version` calls `blob.reload()` for any listed version lacking metadata, and reload won't populate it if the object genuinely has none — one extra API round-trip per such version. Cheap fix: skip reload for blobs that came from `list_blobs` (they already carry metadata).
- Purely numeric `@` suffixes are always interpreted as generations, so a hypothetical all-numeric data version is unreachable. Fine for semver-style versions; maybe worth a sentence in the docs.

Also note the PR is still marked **draft**, so it can't be merged until it's marked ready for review.

vahid-ahmadi (approved):
pls see above

PR #1791, opened 2026-06-24
Add cash ISA and stocks and shares ISA holdings variables

vahid-ahmadi:
## What

Adds two household-level wealth input variables:

- `cash_isa` — amount held in cash ISAs
- `stocks_and_shares_isa` — amount held in stocks & shares (investment) ISAs

Both are `Household`, `STOCK`, `GBP`, GDP-per-capita uprated, mirroring `corporate_wealth` / `savings`.

## Why

The only ISA variable today is the income flow `individual_savings_account_interest_income`. ISA **balances** had nowhere to land:

- investment ISAs were folded into `corporate_wealth` in the (now archived) `policyengine-uk-data` wealth imputation, so they were not separable; and
- cash ISAs were not represented at all.

These variables provide the model-side home for ISA balances imputed by the data pipeline, enabling ISA-balance modelling (ISA reforms, wealth-tax scenarios).

## Data side

The matching data work — surfacing `cash_isa` and `stocks_and_shares_isa` from the Wealth and Assets Survey in the populace UK build — is tracked in PolicyEngine/populace#180.

### DIFF
--- a/changelog.d/add-isa-balance-variables.added.md
+++ b/changelog.d/add-isa-balance-variables.added.md
@@ -1,0 +1,1 @@
+- Add `cash_isa` and `stocks_and_shares_isa` Household wealth input variables for ISA balances.
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: minor
+  changes:
+    added:
+      - Cash ISA and stocks and shares ISA holdings as household wealth input variables.
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,0 @@
-- bump: minor
-  changes:
-    added:
-      - Cash ISA and stocks and shares ISA holdings as household wealth input variables.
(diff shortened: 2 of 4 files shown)


MaxGhenis (approved):
Approving (agent-assisted review). The variables are correct and cleanly scoped: `cash_isa` and `stocks_and_shares_isa` mirror `corporate_wealth`/`savings` exactly (Household, YEAR, float, GBP, OBR per-capita GDP uprating, STOCK), and they're pure inputs with no consumers anywhere in `policyengine_uk/`, so nothing existing moves. No double-count either: `total_wealth` still adds only property/corporate/savings — since the data side folds S&S ISA into `corporate_wealth`, exposing it separately is fine, and cash ISA staying out of `total_wealth` matches its prior (unrepresented) treatment.

One mechanical fix I pushed to the branch rather than bouncing back: the root `changelog_entry.yaml` is inert since the towncrier migration (#1506) — `bump_version.py` and the versioning workflow only read `changelog.d/**`, so merged as-is this would have gotten no version bump and no changelog line. I swapped it for `changelog.d/add-isa-balance-variables.added.md` (the `.added` suffix is what makes the bump come out minor, matching your intent).

Sequencing note for the populace side: populace#181 doesn't depend on this at build time (spec-only), but the eventual UK runtime PR that emits these columns should bump the `uk` extra to the release carrying these variables.

PR #1804, opened 2026-07-25
Format the docs code blocks with current ruff

MaxGhenis:
## Why

Lint fails on `main`. CI runs `uvx ruff format --check .`, which resolves the current ruff release, while contributors format against the version pinned in the dev extras. Four docs pages have drifted out of format, so the check fails on `main` and on every branch cut from it — including unrelated PRs.

## What changed

`uvx ruff format .`, nothing else. Only the Python blocks inside these four pages; no prose or logic changes.

Blocks #1803, which is otherwise green.


MaxGhenis:
Superseded: main is now fully ruff-format clean (1,199 files, including these four docs pages, verified against current origin/main with the same uvx ruff invocation the Lint job uses), so this no longer has anything to do. The Lint failures that motivated it clear on any branch updated past current main.

Issue #1827, opened 2026-08-27
calculate_earnings_quintile ranks the whole population including children, so Q1/Q2 hold no earners


vahid-ahmadi:
## Summary

`dynamics/participation.py::calculate_earnings_quintile` ranks **every person in the dataset, children included**, on actual earnings. Roughly half that population has zero earnings, so the bottom two quintiles are 100% non-earners. Every potential labour-market entrant lands in Q1–Q3, where the OBR Table A1 elasticities are highest — and Q1/Q2 contain no employed wage donors at all, which silently prevents a large share of non-workers from ever entering employment.

## The code

```python
# policyengine_uk/dynamics/participation.py:351-362
employment_income = sim.calculate("employment_income", year)
rng = np.random.RandomState(random_seed)
quintiles = pd.qcut(
    employment_income + rng.random(employment_income.shape),
    q=5, labels=[1,2,3,4,5], duplicates="drop",
)
```

No restriction to adults, no use of potential earnings, and the ranking is **unweighted** over the microdata array, so these are not population quintiles even in principle. (The `+ U(0,1)` noise does break ties, so all five bins are returned — `duplicates="drop"` is not the problem.)

## What it produces

`enhanced_frs_2024_25`, N = 113,617 persons, 50.7% with zero employment income, 20.5% under 18:

| Q | n | share zero-earners | workers | max earnings | share children |
| --- | --- | --- | --- | --- | --- |
| 1 | 22,724 | **100%** | 0 | £0 | 25.2% |
| 2 | 22,723 | **100%** | 0 | £0 | 24.6% |
| 3 | 22,723 | 53.5% | 10,558 | £11,372 | 26.1% |
| 4 | 22,723 | 0% | 22,723 | £29,589 | 25.5% |
| 5 | 22,724 | 0% | 22,724 | £934,873 | 0.9% |

About a quarter of Q1–Q4 are children.

## Two consequences

**1. Elasticity assignment is skewed.** `calculate_participation_elasticities` indexes Table A1 directly by this quintile (`valid_quintiles = np.clip(earnings_quintile[mask] - 1, 0, 4)`), and the table falls steeply as the quintile rises — `lone_parent_3_5` is `[1.554, 1.243, 0.932, 0.621, 0.155]`, a 10× spread. Putting every non-worker in Q1–Q3 places them at the top of that range.

[...]

PR #1828, opened 2026-08-27
Fix hours_worked units in participation dynamics

### LINKED ISSUE #1826: impute_wages_for_nonworkers divides by 52 twice: hours_worked is annual, not weekly
## Summary

`dynamics/participation.py::impute_wages_for_nonworkers` treats `hours_worked` as **weekly** when it is **annual**, dividing by 52 a second time. Imputed earnings for a potential labour-market entrant come out roughly 100× too small, which collapses the extensive-margin response.

## The bug

```python
# policyengine_uk/dynamics/participation.py:216
hourly_wages = np.where(working_mask, employment_income / (hours_worked * 52), 0)
...
imputed_wages[nonworker_mask] = avg_hourly_wage * hours_for_new_entrants * 52
```

The variable's own definition settles the units — `variables/household/income/hours_worked.py`:

```python
class hours_worked(Variable):
    label = "Annual hours worked"
    definition_period = YEAR
    unit = "hour"
```

and the package's own weekly conversion, `variables/household/income/weekly_hours.py`:

```python
    return person("hours_worked", period) / WEEKS_IN_YEAR
```

`dynamics/labour_supply.py:286` reads it the same way:

```python
df["hours_per_week"] = sim.calculate("hours_worked", year) / 52
```

So `participation.py` is the odd one out.

## Empirically

`enhanced_frs_2024_25`, workers with positive income and hours:

| | value |
| --- | --- |
| `hours_worked` percentiles 5/25/50/75/95 | 780 / 1,664 / 1,950 / 2,080 / 2,600 |
| `weekly_hours` median | 37.5 |
| hourly wage as annual (`income / hours`) | **median £15.98, mean £21.14** |
| hourly wage as coded (`income / (hours * 52)`) | median £0.31, mean £0.41 |

2,080 = 40 × 52 exactly. £16 median is consistent with ONS; £0.31 is 3% of the National Living Wage.

Imputed annual earnings from `impute_wages_for_nonworkers`, for entry at 18.8 hours/week: **mean £212, median £237, range £102–£697**, against a correct value of roughly £20,700. (Figures vary a little by simulation year and weighting; the order of magnitude is the point.)

## Impact

Entering work appears to pay nothing. `calculate_gain_to_work` calls this helper with `impute_nonworker_wages=True`, so non-workers' `in_work_income` is affected too, and `imputed_wages` is written straight into `employment_income` via `sim.set_input` at line 531 — there is no downstream conversion that compensates.

In a childcare-reform participation analysis this produced 21 modelled entrants and £0.007 of exchequer revenue from them; with the units corrected, the same code gives 10,031 entrants and £38m.

## Suggested fix

```python
hourly_wages = np.where(working_mask, employment_income / hours_worked, 0)
```

A test asserting the mean implied hourly wage among workers falls in a plausible band (say £8–£40) would have caught this and would catch a future change to the variable's units.

## Related, same units confusion

`participation.py:539` assigns a weekly figure into the annual variable:

```python
new_hours_worked[new_workers] = hours_for_new_entrants   # 18.8
```

so a new entrant is recorded as working 18.8 hours *per year* — `weekly_hours` of 0.36. Should be `hours_for_new_entrants * WEEKS_IN_YEAR`.

Also related: #1827, on `calculate_earnings_quintile` in the same module. The two compound — the empty quintiles are why this helper finds no wage donors for a large share of non-workers.

---

*Correction: an earlier version of this issue named `dynamics/progression.py` as the contradicting file. That was wrong — it does not reference `hours_worked` at all. The contradiction is in `variables/household/income/weekly_hours.py` and `dynamics/labour_supply.py`, as above.*

vahid-ahmadi:
Fixes #1826.

## The bug

`hours_worked` is annual in policyengine-uk — `variables/household/income/hours_worked.py` labels it "Annual hours worked" with a `YEAR` definition period, and `weekly_hours` derives from it as `hours_worked / WEEKS_IN_YEAR`.

`dynamics/participation.py::impute_wages_for_nonworkers` divided by it a second time:

```python
hourly_wages = np.where(working_mask, employment_income / (hours_worked * 52), 0)
```

On `enhanced_frs_2024_25`, among workers with positive earnings and hours:

| | value |
| --- | --- |
| `hours_worked` median | 1,950 (2,080 = 40 × 52 at p75) |
| `weekly_hours` median | 37.5 |
| hourly wage, as annual | **median £15.98, mean £21.14** |
| hourly wage, as coded | median £0.31, mean £0.41 |

£16 is consistent with ONS median hourly earnings; £0.31 is a quarter of the National Living Wage.

## Why it matters

Imputed annual earnings for a non-worker entering work at 18.8 hours a week came out at **£194** against a realistic ~£21,600. Entering work appears to pay almost nothing, so the extensive margin collapses: in a childcare reform analysis built on this module the response was 21 entrants and £0.007 of exchequer revenue from them; with the units corrected the same code gives 10,031 entrants and £38m.

`calculate_gain_to_work` calls this helper with `impute_nonworker_wages=True`, so non-workers' `in_work_income` was affected too, and `imputed_wages` is written straight into `employment_income` via `sim.set_input` — nothing downstream compensates.

## Changes

- Extract the hourly wage calculation into a new `hourly_wage(sim, year)` helper with the correct units, so it can be unit-tested without a dataset.
- `apply_participation_responses` had the same confusion in the other direction, assigning the weekly `hours_for_new_entrants` (18.8) into the annual `hours_worked`, recording a new entrant as working 18.8 hours *per year*. Now annualised.
- Four regression tests pinning the units, using a small fake simulation. They fail on `main` and pass here.

## Not fixed here

The donor pool for this imputation is selected by elasticity group, which is keyed off `calculate_earnings_quintile` — and that function ranks the whole population including children, leaving its bottom two quintiles with no employed donors at all. That is #1827 and is a separate, larger change; this PR is deliberately limited to the units so it can be reviewed and merged on its own.

## Testing

```
policyengine_uk/tests/test_participation_hours_units.py ....                [100%]
4 passed
```

Existing `test_labour_supply_response_formulas.py` and `test_behavioral_responses.py` still pass (6 passed, 11 skipped without a dataset token). `ruff format --check` clean.


juaristi22:
# Adversarial review — policyengine-uk PR 1828

**Verdict: APPROVE WITH COMMENT.**

The two unit fixes are correct. **SHOULD ADDRESS — `test_participation_hours_units.py:50-59`:** the test duplicates the annualization arithmetic but never executes the changed writeback at `participation.py:560-569`. Add an integration-style regression that forces entry and inspects annual `hours_worked`.

Validation: 10 relevant tests passed, 11 credential-dependent tests skipped, Ruff passed, and GitHub checks passed.

juaristi22 (approved):

vahid-ahmadi:
Thanks. Agreed on the gap: the tests pin the arithmetic in `hourly_wage` but never execute the `hours_worked` writeback in `apply_participation_responses`, so that half of the fix is asserted only by inspection.

A direct regression there needs a simulation that will actually produce an entrant — `apply_participation_responses` requires `sim.baseline`, a reform that moves the gain to work, and a stochastic draw to land — so it is a microsimulation-marked test needing dataset credentials rather than the fake-based unit tests in this PR. I would rather add it properly than approximate it with a fake that re-implements the writeback and so tests nothing.

Two options and I am happy either way:

1. Add it here as a `@pytest.mark.microsimulation` test that runs a small reform, forces entry by setting the participation change directly, and asserts `weekly_hours` for new entrants is near 18.8 rather than 0.36.
2. Merge this as the units fix and let #1829 carry it, since that PR changes the same function's donor pool and will need microsimulation coverage anyway.

Given #1829 is stacked on this branch, my inclination is (1) so this PR stands on its own. Say which you prefer and I will push it.

juaristi22:
Let #1829 carry it.

Issue #1832, opened 2026-08-28
childcare_expenses is read as gross cost by Universal Credit and as the parent's net payment by Tax-Free Childcare


vahid-ahmadi:
## Summary

`childcare_expenses` is a single input, documented only as "Total amount spent on childcare". Two programmes read it incompatibly. One of them must be wrong, and which one decides whether Tax-Free Childcare should pay 20% or 25% of it.

Raised by @juaristi22 in review of #1830; filing separately because it is independent of that PR.

## The two readings

**Universal Credit reads it as the total cost of care.** `uc_childcare_element`:

```python
eligible_childcare_expenses = add(benunit, period, ["childcare_expenses"])
covered_expenses = eligible_childcare_expenses * p.coverage_rate   # 0.85
```

UC covers 85% *of childcare costs*, so this is only right if the input is the gross cost.

**Tax-Free Childcare reads it as the parent's own payment.** `tax_free_childcare`:

```python
contribution = (eligible_childcare_expense * p.rate) / (1 - p.rate)   # 0.2/0.8 = 0.25
```

Statutorily, for every £8 a parent pays in, the government adds £2 — so the top-up is **25% of the parent's payment** and **20% of the total**. `rate / (1 - rate)` is therefore only right if the input is the parent's payment net of the top-up.

Both cannot hold for one variable.

## Which is it

I have not been able to settle this from the source data. The FRS `chldcare` block's `chamt` is ingested in `policyengine-uk-data` (`datasets/frs.py`) as:

```python
pe_person["childcare_expenses"] = sum_to_entity(
    childcare.chamt * (childcare.cost == 1) * (childcare.registrd == 1), ...
) * 52
```

with no netting of any support. Variable-level FRS documentation is behind UK Data Service registration and I could not confirm whether respondents are instructed to include money paid out of a Tax-Free Childcare account. @juaristi22 reports that they are, which would make the input **gross** and the TFC formula wrong by a factor of 1.25.

Two further data points, neither decisive:

- `programs.yaml` describes `childcare_expenses` as "out-of-pocket spend net of free hours" in the Childcare Grant entry — net of *free entitlement hours*, which is a different question from net of the TFC top-up.
- DfE's Childcare and Early Years Survey of Parents, the nearest published analogue, asks "amount paid" net of funded entitlement hours but treats reimbursement as a separate question — i.e. gross of support received.

## Why it matters

[...]

PR #1833, opened 2026-08-28
Apply HMRC's payment-active fraction to the Tax-Free Childcare top-up

vahid-ahmadi:
> **Stacked on #1830**, which fixes the pro-rating this builds on. Base is set to that branch and will retarget to `main` when it merges. The diff here is the parameter and its application.

## The problem

`childcare_expenses` is an annual figure, and the model credits every claimant with a full year of it. HMRC's own statistics say otherwise: a claimant's account pays a provider in about **seven months of twelve**.

## Where the number comes from

**Source: [HMRC, Tax-Free Childcare statistics: March 2026](https://www.gov.uk/government/statistics/tax-free-childcare-statistics-march-2026) — Table 2, "Numbers of Children with Open and Used Tax-Free Childcare Accounts and Government Top-up" ([accompanying tables, ODS](https://assets.publishing.service.gov.uk/media/6a0dcb07faac4bc0b0e906ef/Tables_and_Statistics_March_2026.ods)).**

HMRC publishes both a monthly and an annual count of children with **used** accounts, and [defines a used account](https://www.gov.uk/government/statistics/quality-report-tax-free-childcare/tax-free-childcare-quality-report) as one from which a payment was made to a childcare provider in the relevant period. HMRC states explicitly that the annual count is *not* the sum of the monthly counts, because a child appears in every month containing a payment but only once in the year.

The ratio of the two is therefore the mean number of months an account actually pays a provider, as a share of the year:

```
    sum of the twelve monthly counts of children with used accounts
    ---------------------------------------------------------------- ÷ 12
        annual count of children with used accounts
```

| year | sum of monthly | annual unique | months | fraction |
| --- | --- | --- | --- | --- |
| 2018-19 | 1,193,065 | 203,215 | 5.87 | 0.489 |
| 2019-20 | 2,579,660 | 396,365 | 6.51 | 0.542 |
| 2020-21 | 2,610,450 | 461,705 | 5.65 | 0.471 |
| 2021-22 | 4,466,690 | 646,940 | 6.90 | 0.575 |
| 2022-23 | 5,863,685 | 836,220 | 7.01 | 0.584 |
| 2023-24 | 6,833,825 | 966,120 | 7.07 | 0.589 |
| **2024-25** | 7,715,605 | 1,085,020 | **7.11** | **0.593** |
| 2025-26 | 8,042,540 | 1,151,515 | 6.98 | 0.582 |

The 2020-21 dip is the pandemic. The series is otherwise stable at 0.58–0.59 from 2021-22. Each value takes effect from the start of the financial year it measures, so the parameter is a time series rather than a single constant. The full derivation and the table are reproduced in `payment_active_fraction.yaml` alongside the references.

## Applied to the cap as well as the spend

The Tax-Free Childcare cap is statutorily **per entitlement period** (£500 a quarter, £2,000 a year; £1,000 and £4,000 for a disabled child), not an annual allowance. A period without payments forfeits that period's cap rather than carrying it forward, so the fraction applies on the same basis as the spend.

This matters materially. Applied to the spend alone, spending falls only 33% rather than 41%, because the 19% of children at the cap do not move — the correction would be incomplete precisely where awards are largest.

## Effect

`enhanced_frs_2024_25` at 2024, against `policyengine-uk-data`'s own calibration targets:

| | before | after | target |
| --- | --- | --- | --- |
| spending | £1.178bn (1.87×) | **£0.699bn (1.11×)** | £0.63bn |
| caseload | 1,108k (1.02×) | **1,108k (1.02×)** | 1,085k |

Caseload is untouched, as it should be: the fraction changes how many months an account pays, not whether a family claims at all.

That combination is the point. Tax-Free Childcare has been missing its spending target while hitting its caseload target, and no take-up rate can fix that — take-up moves both together, so the rate that lands spending (about 0.43) drops caseload to 0.55× of target. Sweeping the rate on the built dataset:

| take-up | spending | caseload |
| --- | --- | --- |
| 0.586 | 0.98× ✓ | 0.60× ✗ |
| 0.88 (current) | 1.70× ✗ | 0.97× ✓ |

The residual was always the award, not the number of claimants. See PolicyEngine/policyengine-uk-data#470.

An independent check: the correction the model needs is 1 ÷ 1.70 = **0.588**; HMRC's observed fraction is **0.593**. Two unrelated routes — one from the model's error against its targets, one from published administrative data — agreeing to within 1%.

## Caveats, recorded in the parameter file

- HMRC notes the date a payment is made from an account **need not be the date the childcare was received**, so this measures payment activity rather than care received.
- It is a **mean** over a mix of full-year and short-spell claimants. It reproduces the aggregate, not the distribution — adequate for a spending target, and not a basis for claims about which families gain.
- The 0.58–0.64 spread across years and between the children and families constructions is the honest uncertainty; 0.593 is the children figure for the year the targets are set on.

## Testing

```
policyengine-core test policyengine_uk/tests/policy/baseline/gov/hmrc -c policyengine_uk
375 passed
```

Existing expectations in `tax_free_childcare.yaml` are restated at the new values with the arithmetic shown inline, and three tests are added: the fraction reducing a below-cap top-up, pro-rating a cap-bound one, and tracking the year it is measured on (2024 uses 0.593, 2025 uses 0.582). `ruff format` clean.


vahid-ahmadi:
@juaristi22 — review please when you have a moment. This is the follow-up you proposed on PolicyEngine/policyengine-uk-data#472, implemented as a distinct parameter rather than by overloading `tax_free_childcare_eligible_declaration_periods`, per your recommendation.

Two things worth your attention specifically, since both are places I could be wrong:

**The fraction is applied to the cap as well as the spend.** My reasoning is that the Tax-Free Childcare cap is statutorily per entitlement period (£500 a quarter), not an annual allowance, so a quarter without payments forfeits that quarter's cap rather than carrying it forward. This matters materially rather than being a detail: applied to the spend alone, spending falls 33% instead of 41%, because the 19% of children at the cap do not move at all. If you think the cap should not be pro-rated, the result changes from 1.11× to 1.25× and the PR no longer closes the gap.

**I used the children construction, not families.** Your 7.11 months (children, 2024-25) rather than the 7.6 I originally derived from families. I verified it from HMRC Table 2 directly and extended it to a full series — the children and families constructions differ by about 0.05 consistently, and I took children as the right basis for a per-child award. Worth a second opinion.

I have also corrected the record on #472: my `KNOWN_MISSES` removal there was reasoned from a build that reads 1.12×, but that is CI's `TESTING=1` build at 32 calibration epochs rather than 512. The release dataset is 1.87×, so your original framing was right and mine was not.

vahid-ahmadi:
Superseded by a restructured version: the empirical fraction belongs in `policyengine-uk-data`, not as a model parameter. Reasoning and the replacement PR to follow.

juaristi22 (commented):
## Adversarial review — blocking findings

The HMRC count arithmetic is mostly correct, but two source/implementation problems are blocking.

**C1 — Monthly provider-payment activity does not establish quarterly cap usage** (`tax_free_childcare.py:52-70`).

The new ratio measures months in which a payment was made *from* an account to a provider. The statutory cap instead applies to qualifying payments *into* the account during a three-month entitlement period. HMRC’s own TFC35300 example says a £5,000 deposit can leave £3,000 in the excess account, with £2,000 receiving top-up in the next entitlement period. That contradicts the code comment that a period without provider payments forfeits its cap.

Because both `contribution` and `max_amount` are multiplied by `active_fraction`, this is algebraically a uniform award scaler: `min(f*a, f*b) = f*min(a,b)`. It cuts a standard cap-bound 2025 award from £2,000 to £1,164 without a source establishing that 58.2% of quarterly cap entitlement is used.

Please apply the duration proxy to modeled spend only unless a separate quarterly deposit/top-up exposure can be sourced.

Sources:
- [HMRC TFC35200](https://www.gov.uk/hmrc-internal-manuals/tax-free-childcare-technical-manual/tfc35200)
- [HMRC TFC35300](https://www.gov.uk/hmrc-internal-manuals/tax-free-childcare-technical-manual/tfc35300)

**C2 — The launch-year parameter uses the wrong year’s value** (`payment_active_fraction.yaml:3-4`).

HMRC Table 2 gives 2017-18 monthly used-account counts totaling 308,300 and 71,515 annual-unique children: `308,300 / 71,515 / 12 = 0.359249`.

The YAML instead applies `0.489`, the 2018-19 value, from 2017-04-01. That overstates the launch-year factor by about 36%. Add the actual 2017-18 value, start `0.489` in 2018-19, and test both years.

Source: [HMRC March 2026 statistics and Table 2 ODS](https://www.gov.uk/government/statistics/tax-free-childcare-statistics-march-2026).

**Should address**

- HMRC changed the used-account methodology around April 2025. The PR’s 2024-25 ratio combines monthly and annual figures that are not fully definition-consistent. Quantify the effect or use a consistent-year construction. See the [HMRC quality report](https://www.gov.uk/government/statistics/quality-report-tax-free-childcare/tax-free-childcare-quality-report).
- The ratio counts active months but does not weight them by payment value. It is an aggregate duration proxy, not a directly observed share of annual expenditure.
- Re-score after resolving #1832. The reported 1.11×-target result inherits PR 1830’s disputed 25%-of-gross formula, so target agreement is not stable corroboration. Using the PR’s own aggregate figures, I infer that fixing the gross formula and applying activity to spend only is roughly 1.07× target, while also scaling the cap is roughly 0.93×.

Validation: 17 focused and 375 broader HMRC tests passed; parameter metadata and Ruff passed; `git diff --check` passed. GitHub showed no checks attached to head `50bb8dad`.

PR #1846, opened 2026-09-03
Warn instead of silently mislabelling a dataset's fiscal year

### LINKED ISSUE #1845: UKSingleYearDataset silently ignores fiscal_year with file_path, and defaults it to a hardcoded 2025
`UKSingleYearDataset.__init__` (`policyengine_uk/data/dataset_schema.py:42-67`) takes `fiscal_year: int = 2025`, but the argument is only used on the DataFrame branch:

```python
if file_path is not None:
    ...
    self.time_period = str(f["time_period"].iloc[0])   # line 58
else:
    ...
    self.time_period = str(fiscal_year)                # line 67
```

Two problems.

**`fiscal_year` is silently ignored when `file_path` is given.** Passing a conflicting value has no effect and produces no warning:

```
frs_2024_25.h5                        -> time_period = 2024
  same file with fiscal_year=2030     -> time_period = 2024   (argument ignored)
```

**The default is a hardcoded year that drifts.** Building from DataFrames without passing `fiscal_year` silently labels the data 2025:

```
UKSingleYearDataset(person=..., benunit=..., household=...)  -> time_period = 2025
```

The latest FRS release is 2024/25, which on our convention is label 2024 (`CURRENT_FRS_RELEASE.survey_year = 2024`), so the default is already a year ahead of the newest data available, and it will be wrong differently every year. An external collaborator raised exactly this.

Nothing we ship is affected — `create_frs` passes `fiscal_year=year` at all three construction sites — but the failure mode for anyone assembling a dataset by hand is a silently mislabelled `time_period`, which then feeds the wrong policy year into every calculation.

Suggested fix: warn when `fiscal_year` is passed alongside `file_path` (where it cannot take effect), and make `fiscal_year` required on the DataFrame branch rather than defaulting to a hardcoded year.

vahid-ahmadi:
Closes #1845.

`UKSingleYearDataset.__init__` declared `fiscal_year: int = 2025` but used it on only one of its two branches, so both halves failed quietly.

**Ignored alongside a file path.** With `file_path`, the period comes from the file and the argument was discarded without comment:

```
frs_2024_25.h5                     -> time_period = 2024
  same file with fiscal_year=2030  -> time_period = 2024   (silently ignored)
```

**A hardcoded default that drifts.** Building from DataFrames without it labelled the data 2025 — already a year ahead of the newest release, since FRS 2024/25 is label 2024 on our convention, and wrong differently every year:

```
UKSingleYearDataset(person=..., benunit=..., household=...)  -> 2025
```

This came up with an external collaborator who was labelling 2024/25 data as 2025 and reasonably assumed the argument did something.

### This PR

`fiscal_year` now defaults to `None`, and both silent paths warn:

- passing it with `file_path` raises a `UserWarning` saying it cannot take effect, and points at `uprate_dataset` for changing the period;
- omitting it on the DataFrame branch raises a `DeprecationWarning` naming the label it fell back to and restating the convention.

Behaviour is otherwise unchanged: the fallback still yields 2025, kept as `DEFAULT_FISCAL_YEAR`. Nothing we ship is affected — `create_frs` passes `fiscal_year` at all three construction sites.

### Why deprecate rather than require

Making `fiscal_year` mandatory is the better end state, but it breaks any caller relying on the default, so I have not done it here. If you would rather take the break now, say and I will switch the `DeprecationWarning` to a `ValueError` — it is a two-line change and the tests already cover both paths.

### Tests

`test_dataset_fiscal_year.py`: the argument sets the period; omitting it warns and falls back; passing it with a file path warns and the file's period still wins; loading normally warns about nothing.

Full suite: 1,140 yaml policy tests and 181 pytest (44 skipped), ruff clean.

### DIFF
--- a/changelog.d/dataset-fiscal-year-default.md
+++ b/changelog.d/dataset-fiscal-year-default.md
@@ -1,0 +1,1 @@
+- Warned when `UKSingleYearDataset` is given a `fiscal_year` alongside a file path, where it cannot take effect, and when it is built from DataFrames without one.
--- a/policyengine_uk/data/dataset_schema.py
+++ b/policyengine_uk/data/dataset_schema.py
@@ -1,3 +1,4 @@
+import warnings
 import pandas as pd
 import numpy as np
 from typing import TYPE_CHECKING
@@ -40,16 +41,28 @@ class UKSingleYearDataset:
 
         return True
 
+    # Used when a dataset is built from DataFrames without a fiscal year.
+    # Deprecated: pass fiscal_year explicitly, as this will become required.
+    DEFAULT_FISCAL_YEAR = 2025
+
     def __init__(
         self,
         file_path: str = None,
         person: pd.DataFrame = None,
         benunit: pd.DataFrame = None,
         household: pd.DataFrame = None,
-        fiscal_year: int = 2025,
+        fiscal_year: int = None,
     ):
         file_path = str(file_path) if file_path else None
         if file_path is not None:
+            if fiscal_year is not None:
+                warnings.warn(
+                    "fiscal_year is ignored when loading from a file: the "
+                    "time period is read from the file itself. Drop the "
+                    "argument, or use uprate_dataset to change the period.",
+                    UserWarning,
+                    stacklevel=2,
+                )
             self.validate_file_path(file_path)
             with pd.HDFStore(file_path) as f:
                 self._person = f["person"]
@@ -61,6 +74,17 @@ class UKSingleYearDataset:
                 raise ValueError(
                     "Must provide either a file path or all three DataFrames (person, benunit, household)."
                 )
+            if fiscal_year is None:
+                warnings.warn(
+                    "fiscal_year was not given, so this dataset is labelled "
+                    f"{self.DEFAULT_FISCAL_YEAR}. The label is the year the "
+                    "financial year starts, so FRS 2024/25 data is 2024. "
+                    "Pass fiscal_year explicitly; the default will be "
+                    "removed.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+                fiscal_year = self.DEFAULT_FISCAL_YEAR
             self._person = person
             self._benunit = benunit
             self._household = household
--- a/policyengine_uk/tests/test_dataset_fiscal_year.py
+++ b/policyengine_uk/tests/test_dataset_fiscal_year.py
@@ -1,0 +1,66 @@
+"""`fiscal_year` labelling on UKSingleYearDataset.
+
+The year label is the year the financial year starts, so FRS 2024/25 data
+is labelled 2024. A dataset loaded from a file takes its period from the
+file; one built from DataFrames takes it from the argument.
+"""
+
+import warnings
+
+import pandas as pd
+import pytest
+
+from policyengine_uk.data.dataset_schema import UKSingleYearDataset
+
+
+@pytest.fixture
+def frames():
+    person = pd.DataFrame(
+        {"person_id": [1], "person_benunit_id": [1], "person_household_id": [1]}
+    )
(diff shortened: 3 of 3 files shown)


juaristi22:
## Program Review

### Source Documents

- **Pull request**: [#1846 — Warn instead of silently mislabelling a dataset's fiscal year](https://github.com/PolicyEngine/policyengine-uk/pull/1846)
- **PDF**: Not applicable; this is an infrastructure-only dataset-loading change.
- **Year**: N/A
- **Scope**: PR changes only
- **Reviewed head SHA**: `64b1610e3d0de2c7d1c24976b88469bd00d65823`
- **Mode**: full

### Critical (Must Fix)

1. **[C1] Both added warnings are suppressed in ordinary use.** The new calls in [`dataset_schema.py:59`](https://github.com/PolicyEngine/policyengine-uk/blob/64b1610e3d0de2c7d1c24976b88469bd00d65823/policyengine_uk/data/dataset_schema.py#L59-L65) and [`dataset_schema.py:78`](https://github.com/PolicyEngine/policyengine-uk/blob/64b1610e3d0de2c7d1c24976b88469bd00d65823/policyengine_uk/data/dataset_schema.py#L78-L86) cannot reach callers under the package's normal import path. Importing `policyengine_uk` loads LHA modules such as [`LHA_category.py:6`](https://github.com/PolicyEngine/policyengine-uk/blob/64b1610e3d0de2c7d1c24976b88469bd00d65823/policyengine_uk/variables/gov/dwp/LHA_category.py#L6), which installs a process-wide `warnings.filterwarnings("ignore")`; the first active filter becomes `ignore` for every `Warning` subclass. A fresh-interpreter reproduction recorded `default-path=0 file-path=0`, so neither the `DeprecationWarning` nor the `UserWarning` was shown. The new `pytest.warns(...)` contexts force matching warnings visible and therefore pass while masking the integration failure. Scope/remove those global suppressors (and use an end-user-visible category such as `FutureWarning` for the fallback deprecation), then add a fresh-process regression under ordinary filters. Python documents that `DeprecationWarning` is otherwise ignored by default outside `__main__`, while `FutureWarning` targets application end users. ([Python warning categories](https://docs.python.org/3/library/warnings.html))

2. **[C2] The changelog file is invisible to Towncrier.** [`changelog.d/dataset-fiscal-year-default.md:1`](https://github.com/PolicyEngine/policyengine-uk/blob/64b1610e3d0de2c7d1c24976b88469bd00d65823/changelog.d/dataset-fiscal-year-default.md#L1) omits a configured fragment type. Repository guidance requires `changelog.d/{slug}.{added|fixed|changed|...}.md`; running `towncrier build --draft` at the reviewed head produced `No significant changes.` Rename this fix fragment to `changelog.d/dataset-fiscal-year-default.fixed.md` (or another configured type) so release automation consumes it.

### Should Address

None.

### Suggestions

None.

### PDF Audit Summary

| Category | Count |
|----------|-------|
| Confirmed correct | N/A |
| Mismatches (code-path confirmed + visually verified) | N/A |
| Mismatches rejected (code-path cleared) | N/A |
| Unmodeled items | N/A |
| Pre-existing issues | N/A |

### Validation Summary

| Check | Result |
|-------|--------|
| Regulatory Accuracy | Not applicable |
| Reference Quality | Not applicable |
| Code Patterns | 2 critical issues |
| Formatting | Ruff check passed; Ruff format check passed |
| Test Coverage | Focused tests pass, but warning visibility is not tested under ordinary filters |
| Focused Tests | 4 passed |
| PDF Value Audit | Not applicable |
| CI Status | All 6 checks passed, including the 14m11s broad Test check |

### Review Severity: REQUEST_CHANGES

### Next Steps

Resolve C1 and C2, then rerun the fresh-process warning regression, the focused tests, Ruff, Towncrier draft, and the repository test suite. To auto-fix issues: run the fix-pr workflow for this PR.

PR #1844, opened 2026-09-03
Cap LHA at the national maximum and hold frozen rates at the last determination

vahid-ahmadi:
Closes #1842. Closes #1843.

Two defects in the LHA rate, both verified against the gazetted VOA tables.

### 1. The national maximum was not applied

Published rates are `min(BRMA percentile, national maximum)` — Schedule 3B of the Rent Officers (Housing Benefit Functions) Order 1997, published as Table 3 of the VOA workbooks. We applied no cap, so wherever it binds we were far too generous.

Weekly, 2024/25:

| BRMA | cat | published | before | after |
|---|---|---|---|---|
| Central London | B | 331.39 | 455.91 | **331.39** |
| Central London | C | 412.86 | 621.38 | **412.86** |
| Central London | D | 497.10 | 868.77 | **497.10** |
| Central London | E | 704.22 | 1131.63 | **704.22** |
| Inner East London | B | 331.39 | 348.57 | **331.39** |
| Inner East London | D | 497.10 | 526.14 | **497.10** |

Adds `gov.dwp.LHA.maximum` (categories A–E, weekly), sourced from the amending SIs and cross-checked against Table 3 of the 2020/21 through 2026/27 workbooks. Note category A's cap equals category B's and has never bound anywhere.

### 2. Frozen rates were re-based each freeze instead of held

When frozen, `BRMA_LHA_rate` rebuilt the rate from rents indexed to `find_freeze_start`, the first year of the current freeze. For the freeze from April 2025 that meant 2025 rents — but SI 2025/5 and SI 2026/5 carry the April 2024 determination forward, and the published tables are byte-identical across 2024/25, 2025/26 and 2026/27. Every rate was 3.40% high, in every forecast year, since the parameter runs to 2041.

Inner East London category B, annual:

```
          before        after
2024   18,125.64    18,125.64
2025   18,741.84    18,125.64
2026   18,741.84    18,125.64
2027   18,741.84    18,125.64
```

Fixing this needs the parameter re-dated too, not just the lookup. `freeze.yaml` had `2020-01-01: true`, but April 2020 was itself a reset to the 30th percentile with the freeze running from April 2021 — so reading the anchor from the first frozen year was right for 2020–23 only by coincidence. Both changes are needed:

| | 2020–23 | 2025–27 |
|---|---|---|
| old YAML + first frozen year | correct | wrong (anchors 2025) |
| old YAML + last unfrozen year | wrong (anchors 2019) | correct |
[...]

Issue #1842, opened 2026-09-03
LHA rates ignore the national maximum cap, overstating London by up to 81%


vahid-ahmadi:
Published LHA rates are `min(30th percentile of BRMA rents, national maximum cap)` — the cap is Table 3 of the VOA LHA tables. `LHA_cap.py` applies no cap at all, so wherever the cap binds we are far too generous.

Model vs published, weekly, 2024/25:

| BRMA | cat | published | model | error |
|---|---|---|---|---|
| Central London | D | 497.10 | 868.77 | +74.8% |
| Central London | E | 704.22 | 1131.63 | +60.7% |
| Central London | C | 412.86 | 621.38 | +50.5% |
| Central London | B | 331.39 | 455.91 | +37.6% |

The actual September 2023 Central London percentiles were 448.77 / 598.36 / 797.88 / 1150.68 against caps of 331.39 / 412.86 / 497.10 / 704.22 — so the cap does most of the work in London, and we ignore it.

Compounding this, the rent microdata (`parameters/gov/dwp/LHA/lha_list_of_rents.csv.gz`) holds only 2019 and 2020 and is uprated by a single UK-wide private rent index, so local rent divergence since 2020 is lost. That produces large idiosyncratic errors in the other direction:

| BRMA | cat | published | model | error |
|---|---|---|---|---|
| Greater Glasgow | E | 414.25 | 243.35 | −41.3% |
| Greater Glasgow | B | 159.95 | 111.81 | −30.1% |
| Bristol | E | 425.75 | 349.10 | −18.0% |
| Bristol | D | 299.18 | 249.92 | −16.5% |

Mean absolute error across 9 BRMAs x 5 categories for 2024/25 is **15.1%**.

The `A >= B` floor is also missing: Durham publishes category B = category A = 90.50, while the model gives B = 85.49 against A = 74.30.

### Suggested fix

- Add the Table 3 national maximum caps as a parameter and apply them in `LHA_cap` (or in `BRMA_LHA_rate` before the return).
- Apply the `A >= B` floor.
- Longer term, refresh the rents base — a 2019/2020 list scaled by a national index cannot reproduce BRMA-level rates, whatever the cap logic does.

Verified against the VOA `2024-25_LHA_TABLES.xlsx` and the gov.scot LHA rate tables.

PR #1848, opened 2026-09-07
Update PRS rental price index

edward-mcpherson:
Updating PRS uprating parameters from ONS data. Previous series ended in 2024. New data coming in from ONS (here: https://www.ons.gov.uk/economy/inflationandpriceindices/datasets/priceindexofprivaterentsukmonthlypricestatistics) and goes until July 2026. 

---

*Added by @vahid-ahmadi for discoverability — the effect on LHA, since this is more than a data refresh.*

**This is a series swap, not new months on the same measure.** The old values were ONS's Index of Private Housing Rental Prices (Jan 2015 = 100), which was discontinued and replaced by the Price Index of Private Rents (rebased to Jan 2023 = 100). The figures check out as genuine PIPR: July-on-July comes to +3.67% against ONS's reported 3.7% UK rent inflation to July 2026.

**It raises LHA rates about 6.5% from 2024 onward.** The model builds LHA from the 2020 List of Rents scaled by this index, so the 2020→2024 ratio does real work: it goes from 1.14312 to 1.21711. Weekly, uncapped areas:

| BRMA | cat | before | after |
|---|---|---:|---:|
| Maidstone | C | 214.80 | 228.70 |
| Greater Glasgow | B | 111.81 | 119.05 |
| Bristol | C | 217.03 | 231.08 |

2020 itself is unchanged, as that is a reset year taking the 2020 rents directly. Capped areas are unchanged too — Central London B–E stay at the national maximum.

**It moves the model toward the published tables**, so it partly closes the gap left in #1842:

```
Greater Glasgow B   published 159.95 (gov.scot)   before -30.1%   after -25.6%
Bristol C           published 252.00 (VOA)        before -13.9%   after  -8.3%
```

Anyone re-running housing figures should note this lands alongside #1844, which moved rates the other way.


vahid-ahmadi (commented):
Thanks Edward — this is a good catch and the right fix. One thing worth flagging though: it isn't as trivial as the description suggests, and I think the PR should say so.

**It moves LHA rates about 6.5% from 2024 onward.** The model builds LHA from the 2020 List of Rents scaled by this index, so the 2020→2024 ratio is doing real work: it goes from 1.14312 to 1.21711. Measured on current `main`, weekly:

| BRMA | cat | before | after |
|---|---|---:|---:|
| Maidstone | C | 214.80 | 228.70 |
| Greater Glasgow | B | 111.81 | 119.05 |
| Bristol | C | 217.03 | 231.08 |

2020 itself is unchanged, since that's a reset year taking the 2020 rents directly.

**And it moves us toward the published tables**, which is the good part — it partly closes the gap in #1842:

```
Greater Glasgow B   published 159.95   before -30.1%   after -25.6%
Bristol C           published 252.00   before -13.9%   after  -8.3%
```

**Two things for the description rather than the code:**

1. This is a *series swap*, not an extension. The old values are ONS's Index of Private Housing Rental Prices (Jan 2015 = 100); the new ones are the Price Index of Private Rents (2015-01 = 81.3, different base). PIPR is IPHRP's official successor so the swap is correct, but it's a methodology change rather than fresh months on the same measure, and that's worth stating for anyone reading the changelog later.
2. Small typo — the description says the previous series ended in 2014; it ended in January 2024.

Nothing wrong technically as far as I can see. The rebase is harmless because only ratios are used, monthly granularity is preserved through July 2026, and the CPI `uprating` metadata is correctly kept for beyond that.

Happy to approve once the description mentions the LHA effect and the series change — mostly so the ~6.5% movement is discoverable by whoever next wonders why the rates shifted.

vahid-ahmadi:
Completed in #1849

Issue #1854, opened 2026-09-15
Council tax integration: Nuffield Workstream 2 deliverable tracker


vahid-ahmadi:
Tracking issue for the council tax workstream committed in the Nuffield R&D 2025 application (Workstream 2, "Council Tax Integration", £32,000 / 4 person-months, Year 1 Q1 kickoff).

Application text: https://policyengine.github.io/nuffield-rda-2025

## What we promised

> We will add council tax to microsimulation, enabling combined local and national tax burden analysis. This addresses a significant gap—existing models omit local taxation despite £40bn annual revenue.

Method committed (test-driven, citing the childcare precedent):
1. Implement council tax bands and local authority rates from MHCLG data
2. Validation test cases
3. Compare against administrative totals
4. Validate distributional impacts

Deliverables committed:
1. Council tax by band and local authority
2. National model integration
3. Validation report
4. Policy examples

## Current state

More exists than the deliverable list suggests — the **data** layer is substantially built, the **liability calculation** is not.

Already in place:
- `policyengine-uk-data`: `policyengine_uk_data/storage/la_council_tax.csv` — 360 LAs (England, Scotland, Wales) with `band_d_amount`, VOA dwelling counts for bands A–I, and net council tax requirement.
- `policyengine-uk-data`: band assignment to households in `datasets/enhanced_cps.py` (`_assign_council_tax_bands`), using `storage/council_tax_bands_2024.csv` regional shares.
- `policyengine-uk-data`: calibration targets `ons/council_tax_band_d/{code}` and `voa/council_tax/{code}/{band}` emitted from `targets/sources/la_council_tax.py`.
- `policyengine-uk`: the full Council Tax **Reduction** stack under `variables/gov/local_authorities/council_tax_reduction/`, with bespoke schemes for Kingston upon Thames, Merton, Newham, Oxford and Westminster, plus England pensioner / Scotland / Wales parameters.
- `TheAxiomFoundation/rulespec-uk`: the national England pension-age CTR scheme (SI 2012/2885) and 100 local authority CTR schemes encoded with proof citations, plus a 297-row LA registry.

Not in place:
- **Council tax liability is not computed anywhere.** `variables/input/consumption/property/council_tax.py` is an input variable whose own docstring says it is "supplied by the household dataset rather than recomputed from local authority council tax schedules". `council_tax_band` is likewise an input defaulting to Band D.
- No statutory band-ratio parameters (A=6/9 … H=18/9 relative to Band D).
[...]

PR #1855, opened 2026-09-15
Compute gross council tax from LA Band D amounts and statutory band ratios

vahid-ahmadi:
Makes gross council tax a computed quantity (Nuffield deliverable: "Council tax by band and local authority; national model integration").

## Read this first: the aggregate does not mean what it looks like

`local_authority` is **not populated in any dataset**. On both `enhanced_frs_2024_25` and `enhanced_frs_2023_24`, all 52,846 households (31.46m weighted) decode to the enum default, `MAIDSTONE` — 1 distinct LA. So `simulated_council_tax` currently evaluates as "Maidstone's Band D amount (£2,502.59, above the £2,411 England unweighted mean) times the national band distribution", not as an LA-varying quantity.

The LA-varying half of the deliverable is therefore **blocked on policyengine-uk-data assigning `local_authority`**. This PR supplies the model side of it; the data side is the next step.

## What it adds

- `council_tax_band_d_amount` (Household, YEAR, GBP) — reads a new LA-keyed parameter `gov/local_authorities/council_tax/band_d/amounts.yaml`, following the existing `gov/local_authorities/domestic_rates/rates.yaml` precedent for LA-keyed lookups (rather than a dataset input, which would leave the variable at zero everywhere and make the aggregate uncomputable).
  - 344 of 350 LAs with a reported Band D amount are loaded. The 6 unmatched are post-2021 unitaries absent from the `LocalAuthority` enum: North Northamptonshire, West Northamptonshire, Cumberland, Westmorland and Furness, North Yorkshire, Somerset. Northern Ireland has no Band D (domestic rates).
  - Only one year of data is loaded (2026-27 England/Wales, 2025-26 Scotland), so core holds values flat backwards for earlier years. Documented in the variable docstring.
- Band ratio parameters, per country:
  - `england.yaml` — 6:7:8:9:11:13:15:18 relative to Band D = 9 (LGFA 1992 s.5).
  - `wales.yaml` — the same plus band I at 21/9, from 1 April 2005 (Council Tax (Valuation Bands) (Wales) Order 2003). Verified against legislation, not asserted.
  - `scotland.yaml` — A–D unchanged; E–H raised to 473/360, 585/360, 705/360, 882/360 from 1 April 2017 (Council Tax (Substitution of Proportion) (Scotland) Order 2016). **The task brief's flat 9ths would have been wrong for Scotland.**
- `simulated_council_tax` = Band D amount x band ratio, zero for Northern Ireland.

## Design choice (step 3)

**A new variable, `simulated_council_tax`; `council_tax` is left as the dataset input, untouched.** Reasons:

1. `council_tax` feeds `council_tax_less_benefit` and the whole Council Tax Reduction stack. Redefining it would move CTR results for every household in the same commit that introduces the arithmetic — two changes, one diff, no way to attribute a regression.
2. The computed value is not yet comparable: with `local_authority` unpopulated it is a single-LA counterfactual (see above). Switching now would degrade the model.
3. Keeping both lets the comparison be measured. The name follows the repo's existing `simulated_council_tax_reduction_benunit` convention for computed-vs-reported pairs.

Switching `council_tax` over should be a follow-up, once the dataset assigns LAs.

## Aggregate (step 6), 2026, enhanced_frs_2024_25

| | `simulated_council_tax` | `council_tax` (input) |
|---|---|---|
| England | **£61.21bn** | £45.65bn |
| Wales | £4.03bn | £2.84bn |
| Scotland | £7.00bn | £3.97bn |
| UK | £72.24bn | £52.46bn |

Against the ~£40bn England administrative total in the Nuffield application, the computed England figure is **£61.21bn, about 53% too high**. Two known reasons, in order of size:

1. Every household is charged Maidstone's Band D, which is high relative to the England mean (the single-LA artefact above).
2. This is **gross** liability: no single-person discount, no exemptions, no CTR. Those alone are worth many billions.

So this number is not a validation of the arithmetic — it is a measurement of how far the missing pieces reach. On `enhanced_frs_2023_24` the England figure is £62.72bn against a £53.82bn input, same pattern.

## Dangling variable in policyengine-uk-data — confirmed, and a hazard this PR creates

`policyengine_uk_data/targets/sources/la_council_tax.py:182` emits targets with `variable="council_tax_band_d_amount"`. Confirmed: `git grep council_tax_band_d_amount` in policyengine-uk returned nothing before this PR.

**It is silently skipped today, not raised.** Traced: `build_loss_matrix.py::_compute_column` has no route for the `ons/council_tax_band_d/` prefix (it routes `voa/council_tax/` and `obr/council_tax`), so it falls through to `_compute_simple_gbp`, which does `ctx.sim.tax_benefit_system.variables[variable]` inside a `try/except KeyError` and returns `None`; the caller then hits `if col is None: continue`. Not even the `logger.debug("No compute logic...")` line fires. No warning is emitted anywhere.

**Hazard:** once this merges and policyengine-uk-data bumps its policyengine-uk pin, that target becomes live — and `_compute_simple_gbp` **sums** `council_tax_band_d_amount` across the LA's households, while the target value is a single Band D **level** (~£2.5k). Sum-vs-level; it will pull calibration the wrong way. **Do not bump the pin until `la_council_tax.py:182` is changed to a weighted mean / custom compute, or dropped.** Not fixed here — different repo.

## Tests

12 new YAML cases in `tests/policy/baseline/gov/local_authorities/council_tax/`: all eight England band ratios in a named LA (Maidstone), the Wales-only band I in Cardiff, Scotland band H in Edinburgh, and Band D identity in both Maidstone and Cardiff.

Actual output:
- `policyengine-core test policyengine_uk/tests/policy`: **1130 passed**
- `pytest policyengine_uk/tests/`: **149 passed, 43 skipped** (96s)

Nothing failed, so no base-commit comparison was needed. Note: both were run with `PYTHONPATH=.` — the local environment has a stale non-editable `policyengine_uk` in site-packages that the `policyengine-core` CLI otherwise picks up in preference to the working tree.

## Not covered

- **LA assignment in the dataset** — the biggest gap; see the top of this description.
- Single-person 25% discount, and all other discounts.
- Exemptions (student households, Class A–W).
- Empty-home and second-home premiums.
- Disabled band reduction.
- Scotland and Wales liability beyond the band-ratio arithmetic — no water charges (Scotland), no country-specific reliefs.
- Precepts (parish, police, fire) are inside the published Band D figures but are not modelled separately or made reformable.
- `council_tax` itself is unchanged, so CTR results are unaffected by this PR.
- Band D amounts are a single year held flat backwards; no historical series, no uprating.


vahid-ahmadi:
Flagging a design question for review that I think is the crux of this PR, plus a downstream hazard it creates.

## The £61.21bn is an artefact, and the cause may be architectural rather than a missing dataset field

This PR computes liability as `council_tax_band_d_amount` (an LA-keyed parameter, looked up from `household("local_authority")`) × band ratio. The England aggregate comes out at **£61.21bn against the ~£40bn** administrative total cited in the Nuffield application.

The PR attributes this partly to `local_authority` being unpopulated — every household decodes to the enum default `MAIDSTONE`, so the result is "Maidstone's Band D (£2,502.59, above the £2,411 England unweighted mean) × the national band distribution".

That observation is correct, but I do not think the conclusion "the dataset needs to assign LAs" follows. Checking `policyengine-uk-data`:

- No dataset builder ever writes a `local_authority` value onto households — `grep local_authority policyengine_uk_data/datasets/*.py` returns only target-matrix plumbing.
- LA analysis is instead done by **reweighting the national sample per LA**: `create_local_authority_target_matrix` in `datasets/local_areas/local_authorities/loss.py` builds a target matrix over `local_authorities_2021.csv` codes, and `calibrate.py` writes one weight vector per LA to `local_authority_weights.h5`.

So `local_authority` appears to be unpopulated **because per-household LA tagging is not the mechanism**, not because of an oversight. If that reading is right, then a household-level variable that looks up a parameter by `household("local_authority")` cannot work under the current architecture — it will return the default LA's amount nationally, which is exactly the overstatement observed.

**Question for review:** should this be computed per-LA under the LA weighting context instead of via a household-level enum lookup? I would rather have that ruled on before the approach is built further, since the remaining deliverables (validation against administrative totals, distributional impacts) all depend on it.

Worth noting the £61.21bn also includes no single-person discount, exemptions, premiums or CTR, all of which are declared gaps here — so some overstatement is expected regardless. But the single-LA artefact is the larger term and it is the one with a design question attached.

## Downstream hazard: filed as PolicyEngine/policyengine-uk-data#483

Independent of the above — merging this PR makes a latent defect in `policyengine-uk-data` go live.

That repo emits a target at `targets/sources/la_council_tax.py:182` with `variable="council_tax_band_d_amount"`. Today that variable does not exist, so `_compute_simple_gbp` hits `KeyError` → returns `None` → the target is silently discarded (not even the `logger.debug` fires). Once this PR merges and the pin is bumped, the target activates — and `_compute_simple_gbp` **sums** the variable across households while the target value is a single Band D **level** of ~£2,500. Weighted sum (order 1e10) compared against 2.5e3.

It will not raise. It will just calibrate against a wrong comparison. Detail and options in #483; that needs resolving before the pin is bumped, not before this merges.

## Verified separately

The Scotland correction in this PR is right and worth not losing: Scotland raised the E–H multipliers to 473/360, 585/360, 705/360, 882/360 from April 2017 (SSI 2016/123), so the flat 9ths only hold for England. Wales Band I = 21/9 traces to LGFA 1992 s.5(1A) itself (proportions 6:7:8:9:11:13:15:18:21 for Wales from April 2005), not to separate Welsh legislation.

Also relevant to the Wales lane, from the parallel Axiom review (TheAxiomFoundation/rulespec-uk#344): **LGFA 1992 s.12 was omitted with effect 1 April 2026** by the Local Government Finance (Wales) Act 2024 (asc 6), ss.18(6) and 23(7) — and 2026-27 is the year this PR's Band D amounts are drawn from.

Context and full deliverable scope: #1854.

juaristi22 (changes requested):
The arithmetic is right, and both new variables have formulas, so they are safe for the Microcosm adapter. Requesting changes on data coverage, not on the design.

**On the design question.** The answer differs by consumer. policyengine-uk-data reweights one national sample per authority, so a household-level enum can never vary by authority there. Microcosm assigns an authority to every household row (an ONS code from its geography ladder), so a household-level `local_authority` lookup is exactly the right shape for us. What is missing is on our side: nothing maps that code into the `LocalAuthority` enum yet, so Microcosm builds also show the Maidstone default. We will supply `local_authority` as an input from the dataset build. The same gap already silences the bespoke CTR schemes in `council_tax_reduction/config.py` (Merton, Kingston upon Thames, Newham and the others test `local_authority ==`) on every dataset, which raises the priority of populating the input.

Changes requested:
1. **Enum coverage.** Add the six 2023 unitaries (North Northamptonshire, West Northamptonshire, Cumberland, Westmorland and Furness, North Yorkshire, Somerset) to `LocalAuthority`. Microcosm's roster is the April 2023 local authority list, so those six are exactly the codes that would fall to the default, and adding them lets all 350 Band D rows load.
2. **Vintages.** The parameter carries only 2026-27 for England and Wales and 2025-26 for Scotland. Microcosm measures at 2025, so England and Wales would be charged next year's level. The MHCLG Table 10 workbook prints the previous-year column alongside the current one, so fiscal 2025 values are in the same file, and Scotland's 2026-27 average Band D by council is published in the Scottish Government council tax datasets. Please carry 2025 and 2026 for all three nations.
3. **References.** The England `amounts.yaml` href is the statistics collection page. Cite the specific table (Table 10, local authority level data, 2026-27) and say which Band D basis it is (including all precepts, if it is the area figure). The Scotland href should name the specific dataset.
4. **Scotland ratios before 2017.** `scotland.yaml` starts every band at 2017-04-01, including A to D, which SSI 2016/123 did not change. Date A to D from 1993-04-01 so the parameter is defined before 2017.

Agreed on the hazard: policyengine-uk-data must merge PolicyEngine/policyengine-uk-data#484 before bumping its pin past this PR. Microcosm has no exposure; the Band D rate was never ported (it is a signed exclusion in our parity register).

PR #1864, opened 2026-09-18
Add statutory council tax discounts (LGFA 1992 s.11, England)

vahid-ahmadi:
Adds the statutory council tax discounts under section 11 of the Local Government Finance Act 1992 (England). Nuffield Workstream 2 deliverable; part of #1854.

## What this adds

- **Parameter** `gov.local_authorities.council_tax.discounts.appropriate_percentage` — 25% from 1993-04-01, referenced to LGFA 1992 s.11(3).
- **Variable** `council_tax_discount_rate` (household) — s.11(1) applies the appropriate percentage where there is one resident who is not disregarded; s.11(2) applies **twice** the appropriate percentage where there is no such resident.
- **Variable** `simulated_council_tax_less_discounts` (household) — `simulated_council_tax * (1 - council_tax_discount_rate)`.
- YAML policy tests plus a pytest that pins the derivation.

The existing `council_tax` input variable is untouched.

## The 25% is the parameter; 50% is derived

s.11(3) sets the appropriate percentage at 25 per cent "unless the Secretary of State by order otherwise provides". The familiar 50% for an unoccupied dwelling is **not a printed statutory figure** — it is s.11(2)'s "twice the appropriate percentage". So only 0.25 is encoded, and the no-resident rate is computed as `2 * p`.

`policyengine_uk/tests/test_council_tax_discount_derivation.py` proves this: it sets the parameter to 0.3 and asserts the single-resident rate becomes 0.3 **and** the no-resident rate becomes 0.6, plus a p=0 case that zeroes both. A hardcoded 0.5 would fail these. This test is a pytest rather than a YAML test because the YAML test runner's `reforms:` key takes reform module paths, not parameter overrides.

## Disregards: what is exact and what is not

LGFA 1992 **s.6(5)** defines a resident as an individual aged 18 or over who has their sole or main residence in the dwelling. So counting residents as adults aged 18+ **is** the statutory test, not an approximation.

The approximation is narrower and is this: **Schedule 1 disregarded-person categories are not modelled**, so every resident is treated as never disregarded. Consequences, stated in the variable docstring too:

- A two-adult household where one adult is disregarded (e.g. a full-time student) *should* get the 25% discount. Here it gets **none**.
- A household whose only adults are all disregarded *should* get 50%. Here it gets **none**.

The discount is therefore **understated**, not overstated. Adding Schedule 1 will change these results.

## Relationship to #1855

This is based on `council-tax-computed-liability` (#1855), **not** main, and the PR base is set accordingly. This is a genuine dependency, flagged rather than stacked silently:

- Applying a discount requires a gross-of-discount base. On `main` the only council tax figure is the FRS-supplied `council_tax` input. pe-uk-data documents that variable as "the gross liability before CTR" (`targets/compute/council_tax.py`) — i.e. gross of Council Tax Reduction, and silent on statutory discounts. FRS-reported council tax is generally the amount the household is billed, which is already net of any single-person discount; multiplying it by `(1 - rate)` would double-discount. I am flagging that FRS point as inference, not as something the repo states.
- `simulated_council_tax` from #1855 is gross of discounts by construction, so it is the correct base.

**The parameter and `council_tax_discount_rate` are fully independent of #1855's open design question** (whether council tax should be computed under the per-LA weighting context rather than a `household("local_authority")` enum lookup). They are driven by household composition only and survive either resolution. If #1855 changes shape, only `simulated_council_tax_less_discounts` needs re-pointing at whatever the gross variable becomes. Nothing here touches the LA lookup design.

## Microcosm adapter safety — safe

Checked `packages/microcosm-frame/src/microcosm/frame/adapters/policyengine_uk.py`. `_is_engine_computed` (lines 157–162) returns True if a variable has `adds`, `subtracts`, `formula`, or `formulas`. Both new variables define a `formula`, so both classify as engine-computed and are materialised by `materialize()` via `simulation.calculate`. Neither can ever be requested as a supplied input column. The new parameter is not a variable and is invisible to the adapter. No incompatibility found.

## Test output

All run with `PYTHONPATH=.` (a stale non-editable `policyengine_uk` in site-packages otherwise shadows the working tree).

`policyengine-core test policyengine_uk/tests/policy -c policyengine_uk`:
```
1135 passed, 1 warning in 23.25s
```

Council tax directory only (5 new cases + 12 existing from #1855):
```
17 passed, 1 warning in 0.21s
```

`pytest policyengine_uk/tests/ --ignore=policyengine_uk/tests/microsimulation`:
```
152 passed, 17 skipped, 6 warnings in 93.38s
```

Derivation test alone:
```
3 passed, 1 warning in 5.03s
```

The `microsimulation/` directory was excluded as it requires dataset downloads; no source there is touched by this change.

## Out of scope — follow-ups

Deliberately not built here:

- **Schedule 1** disregarded-person categories (students, severe mental impairment, carers, apprentices, etc.) — the largest gap, and the one that changes numbers above.
- **s.11A** billing-authority determinations (discounts varied locally for empty/second homes).
- **s.11B / s.11C** long-term-empty and second-home premiums.
- **s.11D** excepted classes of dwelling.
- **Exemptions** (Class A–W exempt dwellings).
- **Wales** — Part I also applies to Wales, but ss.12, 12A–12B give Welsh Ministers separate powers that have not been verified here.
- **Scotland** — separate provision under s.79.

The discount returns zero outside England.

Closes part of #1854.


juaristi22 (commented):
The encoding is right: 0.25 as the parameter, the no-resident rate derived as twice it, and the pytest pins the derivation. Not blocking while this sits on #1855; two asks before approval.

1. **Scope across nations.** The rate is zero outside England, so `simulated_council_tax_less_discounts` is net of the discount in England and gross in Wales and Scotland. Both apply a 25 per cent single-person discount in practice (Scotland under section 79 of the same Act). Either apply it GB-wide with the citations, or name the variable as England-only, so that a UK aggregate does not mix bases. This matters for the validation step in #1854 and for Microcosm's OBR nation rows.
2. **Zero-adult branch.** On survey households the no-resident case is unreachable (FRS households have an adult reference person), so the doubled rate only ever fires on synthetic inputs. Fine as statute; say so in the docstring.

Minor: the resident count via `is_adult` (age 18 or over) matches the FRS `adulth` single-adult flag that both policyengine-uk-data and Microcosm use for council tax cell means, so the two sides agree on who counts.

Nothing here needs porting to Microcosm. Once #1855 lands and Microcosm supplies `local_authority`, the CTB line 8 single-adult-discount dwelling counts per authority (already in Chronicle) become a validation surface for this variable, since counts are additive.

